PR #1, opened 2019-10-01
OM-44 - Create initial UI

jimppa:


.eslintrc.json
```diff
@@ -0,0 +1,11 @@
+{
+  "extends": ["react-app", "plugin:@typescript-eslint/recommended"],
+  "overrides": [
+    {
+      "files": ["**/*.tsx"],
```
midori0507:
```suggestion
      "files": ["**/*.tsx", "**/*.ts"],
```

.gitignore
```diff
@@ -0,0 +1,23 @@
+# See https://help.github.com/articles/ignoring-files/ for more about ignoring files.
+
+# dependencies
+/node_modules
```
midori0507: Without `/` maybe?

src/i18n/i18n.ts
```diff
@@ -0,0 +1,29 @@
+import i18n from 'i18next';
```
midori0507: rename to i18nInit?

jimppa: I also first thought about init. However the default export from this file `i18n`-instance/promise is passed to the provider in BrowserApp. What do you think?

.gitignore
jimppa: This is what CRA does by default. What is the difference, I have no idea?

midori0507: Ah so nvm then!

src/i18n/i18n.ts
midori0507: You can name them as anything when you import it, default export is flexible
`import i18n from ./i18nInit.ts`

PR #6, opened 2019-10-16
OM-45 tunnistamo login

jimppa:
Add redux-oidc (plus dependencies) for authenticating with tunnistamo.


src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+import { Provider as ReduxProvider } from 'react-redux';
+import { OidcProvider, loadUser } from 'redux-oidc';
 
+import userManager from './oidc/userManager';
+import enableOidcLogging from './oidc/enableOidcLogging';
 import Home from './pages/Home';
+import OidcCallback from './pages/OidcCallback';
 
-const App: React.FC = () => {
+if (process.env.NODE_ENV !== 'production') {
```
midori0507: Is this somehow temporary?

src/BrowserApp.tsx
```diff
@@ -1,15 +1,27 @@
...
 import { BrowserRouter } from 'react-router-dom';
 import { I18nextProvider } from 'react-i18next';
+import { getDefaultMiddleware } from 'redux-starter-kit';
+import { USER_FOUND } from 'redux-oidc';
 
 import App from './App';
 import i18n from './i18n/i18nInit';
+import createStore from './redux/store';
+
+const store = createStore({
```
midori0507: I think you should bring this part to the store file,  instead defining it here and pass as prop.

src/App.tsx
```diff
@@ -1,14 +1,34 @@
...
+  enableOidcLogging();
+}
+
+type Props = {
+  store: Store;
+};
+
+function App(props: Props) {
```
midori0507: and then invoke to get the store here?

src/App.tsx
jimppa: oidc-client does not log anything by default, I just enabled logging (to console) that could help when developing (e.g. user logged in, token expired.. might be helpful)

PR #10, opened 2019-10-18
OM-115 graphql client

jimppa:
Using urql as client


midori0507:
Getting back to `apollo-client`, reactive programming is not for everybody

jimppa:
We'll use apollo-client instead.

PR #11, opened 2019-10-23
OM-115 | Apollo client

jimppa:


.eslintrc.json
```diff
@@ -13,6 +13,7 @@
     "@typescript-eslint/func-call-spacing": ["error"],
     "@typescript-eslint/member-ordering": ["warn"],
     "@typescript-eslint/no-require-imports": ["error"],
+    "@typescript-eslint/camelcase": "off",
```
midori0507: I think it shouldnt off..

src/oidc/userManager.ts
```diff
@@ -5,15 +5,13 @@ const location = `${window.location.protocol}//${window.location.hostname}${
   window.location.port ? `:${window.location.port}` : ''
 }`;
 
-/* eslint-disable @typescript-eslint/camelcase */
```
midori0507: put it back and enable the rule? That rule is quite needed

PR #21, opened 2019-12-13
Om 254 add mobile layout support

tepazuu:


src/common/layout/PageLayout.module.css
```diff
@@ -15,4 +23,27 @@ header {
...
+    width: 400px;
+}
+
+@media (max-width: 1200px) {
+    button {
+        width: 300px;
+    }
+  }
+
+  @media (max-width: 320px) {
```
jimppa: Since 414px is used in other places for the smallest breakpoint, I would suggest you use it here too for consistency.

src/common/header/Header.module.css
```diff
@@ -2,10 +2,23 @@
...
+  padding: 10px 0;
 }
 
-.appName {
-  font-size: 24px;
-  margin-left: 30px;
-}
\ No newline at end of file
+
+  .appName {
```
jimppa: Make these styles mobile-first. To do so, copy the styles for the **smallest** screen here (mobile-first!), then start applying styles that kick into action **after** a certain width, avoid max-width media queries. Like this:

Mobile:
```
.appName {
  font-size: 16px;
  margin-left: 0;
}
```

Next smallest screen:
```
@media (min-width: 415px) {
  .appName {
    font-size: 24px;
    font-weight: bold;
    margin-left: 10px;
  }
}
```

PR #28, opened 2020-01-10
OM - 245 | Submit approved registration

tuovinensanttu:
Profile approval works and redirects user to confirmation screen.
Improved translations.


src/pages/memberhsip/helpers/convertBooleantoString.ts
```diff
@@ -0,0 +1,8 @@
+export default function convertBooleanToString(
```
jimppa: Could you add some basic tests for this?

tuovinensanttu: Added tests for all existing helper functions.

PR #38, opened 2020-02-04
Om - 421 | Send language at home

tuovinensanttu:
- Removed comments from langaugeAtHome field
- Changed photoUsageApproved value to string inside form.

### DIFF
--- a/src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
+++ b/src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
@@ -140,7 +140,7 @@ function CreateYouthProfileForm(props: Props) {
         schoolName: '',
         schoolClass: '',
         language: '',
-        photoUsageApproved: 'false',
+        photoUsageApproved: false,
         approverFirstName: '',
         approverLastName: '',
         approverEmail: '',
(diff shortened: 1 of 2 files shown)


src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
```diff
@@ -396,7 +399,7 @@ function CreateYouthProfileForm(props: Props) {
                         id="photoUsageApprovedYes"
                         name="photoUsageApproved"
                         type="radio"
-                        value={true}
+                        value={'true'}
```
pjlampil: Does this belong here?

src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
```diff
@@ -409,7 +412,7 @@ function CreateYouthProfileForm(props: Props) {
                         id="pphotoUsageApprovedNo"
                         name="photoUsageApproved"
                         type="radio"
-                        value={false}
+                        value={'false'}
```
pjlampil: Or this?

tuovinensanttu: Technically no. Photo usage radio button was broken, so I though I give it a quick fix. Now that I think of it, the fix is kinda unnecessary, because the value isn't even send to backend at the moment. I can make these changes at later date.

PR #46, opened 2020-02-19
OM - 490 | Logout functionality

tuovinensanttu:
Added logic that allows user to log out. Because the design included additions to header & dropdown functionality, I added those as well.

Redirecting back to application after logout isn't working, that's a feature in current version of Tunnistamo.


src/common/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+      {isDropdown && (
+        <React.Fragment>
+          <button
+            className={styles.navButton}
+            onClick={() => toggleDropdown(prevState => !prevState)}
+          >
+            {navBarItem.label}
+            <img
+              src={navBarItem.icon}
+              alt={t('nav.menuButtonLabel')}
```
Jokinen: Consider whether you really need the `alt` attribute here. When the image is right next to a text that describes its purpose, you don't necessarily need to label the image separately. You do need the alt text if the icon is more than just decorative. If you use this element with a screen reader it would read out `navBarItem.label` and then `t('nav.menuButtonLabel')` right after. Would that be valuable?

To avoid linting errors, you must use `alt=""`.

> This means that the alt attribute (sometimes called the alt tag, though technically this is incorrect) is not the only mechanism for providing the content and function of the image. This information can also be provided in text adjacent to the image or within the page containing the image. In some cases where the equivalent cannot be presented succinctly, a link to a separate page that contains a longer description of the image content can be provided.

https://webaim.org/techniques/alttext/

If the icon is more than just decorative, consider whether the alt text should be changed from menu to something else.

src/common/dropdown/Dropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+      document.removeEventListener('click', handleClick);
+    };
+  }, []);
+
+  return (
+    <div className={styles.dropdownWrapper} ref={ref}>
+      {/* The button text is the first item in the option list */}
+      {isDropdown && (
+        <React.Fragment>
+          <button
```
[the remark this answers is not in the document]
Jokinen: You should also see about adding a focused state to button within the dropdown component.

src/common/dropdown/Dropdown.tsx
Jokinen: I tested the menu out with voice over, and currently it reads (when logged in).

"FirstName LastName Menu"

I think the icon does have more than just a decorative purpose here. It denotes--to the seeing user--that this is a profile menu.

I think that the most correct way to resolve the issue would be to change the alt text for the icon to be "profile menu". This would meant that you should probably pass it as a prop for the `dropdown` component instead of declaring it within the component itself.

src/common/dropdown/Dropdown.tsx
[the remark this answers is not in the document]
tuovinensanttu: Added `aria-haspopup` and `aria-expanded`. Buttons have outline when focused.

src/common/dropdown/Dropdown.tsx
tuovinensanttu: Alt text is now passed to dropdown and its value is changed to "Profile menu".

src/common/header/userDropdown/UserDropdown.tsx
```diff
@@ -0,0 +1,102 @@
...
+  const empty = {
+    id: 'emptyId',
+    label: ' ',
+  };
+
+  const dropdownOptions = getDropdownOptions();
+
+  return (
+    <React.Fragment>
+      <Dropdown options={dropdownOptions} />
```
Jokinen: Did you forget to pass the `alt` text here?

PR #47, opened 2020-02-21
OM - 478 | User can skip to registration form

tuovinensanttu:
If birthDate is missing from cookies, user will always be redirected to login page. BirthDate value is reset to empty during logout process. This will help in cases where two people wan't to register one after another using same computer.

### DIFF
--- a/src/auth/logout.ts
+++ b/src/auth/logout.ts
@@ -3,7 +3,7 @@ import store from '../redux/store';
 import { apiError } from './redux';
 
 export default function(): void {
-  // Reset birthDate cookie here as well. This way we will never run to a problem
+  // Reset birthDate cookie here. This way we will never run to a problem
   // where user is redirected to registration from without or with wrong birthDate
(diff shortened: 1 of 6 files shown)


src/auth/logout.ts
```diff
@@ -3,6 +3,9 @@ import store from '../redux/store';
 import { apiError } from './redux';
 
 export default function(): void {
+  // Reset birthDate cookie here. This way we will never run to a problem
+  // where user is redirected to registration from without or with wrong birthDate
+  document.cookie = 'birthDate=';
```
Jokinen: Cookies are not my forte, but what happens to other cookies when you set this?

tuovinensanttu: They are still there, document.cookie looks like a normal string but it behaves like name-value pair. When setting `document.cookie  =` you would think it replaces the whole cookie string, but instead just adds the value to existing string.

PR #48, opened 2020-02-21
OM - 499 | Approver fields are no longer required if user is adult.

tuovinensanttu:
Changed validation rules so if user is adult, approver fields are still shown but they are not required. Also changed hard coded number values to constants.


src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
```diff
@@ -34,20 +39,36 @@ const schema = Yup.object().shape({
...
-    .min(2, 'validation.tooShort')
-    .max(255, 'validation.tooLong')
-    .required('validation.required'),
-  approverPhone: Yup.string()
-    .min(6, 'validation.phoneMin')
-    .required('validation.required'),
-  approverEmail: Yup.string()
-    .required('validation.required')
-    .email('validation.email'),
+  approverFirstName: Yup.string().when([], () => {
```
Jokinen: The `callback` function for `when` has two parameters: `(value, schema) => ...`. Instead of declaring `Yup.string()` later on, you could just "extend" the `schema` you have already created.

Not actually referencing a value with `when` seems like the most pragmatic solution here. But I'll mention here that `yup` does have mechanism for injecting values that don't exist within the validated object: `context`. This may be useful if you have to use similar logic with "runtime" values. Formik doesn't support it our of the box though, so I think that the current approach is the most correct.

The `validate` function looks to have been declared appropriately in this example for context usage:
https://codesandbox.io/s/oqvp75vxxz

Jokinen:
Small style comment, but otherwise the code looked clean 👍

src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
tuovinensanttu: The reason I went with `Yup.string()` was because I couldn't figure out what type to give to schema object (I was trying to avoid using "any" and then suppressing the warning). Now I finally figured out appropriate type for it and went with the approach you suggested.

PR #51, opened 2020-02-26
OM - 498 | Add back button 

tuovinensanttu:
Added back button for under 13 years old view. Now user can return to view that asks birthdate without reloading page.

### DIFF
--- a/src/auth/components/login/Login.module.css
+++ b/src/auth/components/login/Login.module.css
@@ -10,14 +10,6 @@
   }
 }
 
-.linkArrow{
-  color: #0072c6;
-}
-
-.loginHelpText:hover {
-  cursor:pointer;
-}
-
 .serviceLink {
   background-color: var(--hds-ui-color-black);
   color: var(--hds-ui-color-white);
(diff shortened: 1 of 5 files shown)


src/auth/components/login/Login.module.css
```diff
@@ -25,4 +25,13 @@
...
+  margin-bottom: 30px;
+}
+
+.goBack {
+  text-decoration: underline;
+}
+
+.goBack:hover {
+  cursor: pointer;
 }
```
Nizar-Rahme:
```suggestion
.goBack {
  text-decoration: underline;
  cursor: pointer;
}
```
No need for `hover` pseudo-class here.

src/auth/components/login/Login.tsx
```diff
@@ -62,6 +62,15 @@ function Login(props: Props) {
...
               {t('login.findNearestService')}
             </a>
+            <p>
+              <span
+                role="button"
+                onClick={() => setShowManualRegistration(false)}
+                className={styles.goBack}
+              >
+                {t('login.return')}
+              </span>
```
Nizar-Rahme: Any reason why you opted for `span` instead of `button`?

src/auth/components/login/Login.module.css
tuovinensanttu: I ended up removing the whole class.

PR #52, opened 2020-02-27
OM - 494 | Update view after renew

tuovinensanttu:
Includes tickets 
- OM - 495
- OM - 492

Renew button is shown based on value returned from backend. Successful renew shows notification and refreshes expiration date. To test this functionality, change expiration date in django admin to point year, 2019 for example.

### Other changes
NotificationComponent 
- Added style to close button so it doesn't appear on top of the title text.
- Changed import method for DismissableNotification. Previous method `import { DismissableNotification } from 'hds-react'` would crash during testing.

Tests
- Added tests for MembershipInformation component
- This required installing extra package to help jest render component with canvas.

### DIFF
--- a/src/common/notification/NotificationComponent.module.css
+++ b/src/common/notification/NotificationComponent.module.css
@@ -9,5 +9,14 @@
     margin: 0;
 }
 
+/*
+    Use !important for now to shrink notification button size
+    and prevent button appearing on top of the title text.
+    Replace after DismissableNotification accepts className as prop.
+*/
+.notification button {
+    width: 50px !important;
+}
+
 
 
--- a/src/common/test/testUtils.tsx
+++ b/src/common/test/testUtils.tsx
@@ -1,6 +1,7 @@
 import React, { ReactElement } from 'react';
 import { Provider } from 'react-redux';
 import { mount, shallow } from 'enzyme';
+import { act } from 'react-dom/test-utils';
 
 import store from '../../redux/store';
 
@@ -9,3 +10,11 @@ export const mountWithProvider = (children: ReactElement) =>
 
 export const shallowWithProvider = (children: ReactElement) =>
   shallow(<Provider store={store}>{children}</Provider>);
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+export const updateWrapper = async (wrapper: any) => {
+  await act(async () => {
+    await new Promise(resolve => setTimeout(resolve, 0));
+    wrapper.update();
+  });
+};
--- a/src/pages/membership/components/membershipInformation/MembershipInformation.tsx
+++ b/src/pages/membership/components/membershipInformation/MembershipInformation.tsx
@@ -24,6 +24,7 @@ const RENEW_MEMBERSHIP = loader('../../graphql/RenewMyYouthProfile.graphql');
 type Props = {
   expirationDate: string;
   status: MembershipStatus | undefined | null;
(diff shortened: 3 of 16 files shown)


src/pages/membership/components/membershipInformation/MembershipInformation.tsx
```diff
@@ -11,41 +11,49 @@ import NotificationComponent from '../../../../common/notification/NotificationC
...
       input: {},
     };
 
-    renewMembership({ variables }).catch(() => setShowNotification(true));
+    renewMembership({ variables })
+      .then(result => {
+        if (result.data) {
+          setSuccessNotification(true);
+        }
+      })
```
Nizar-Rahme:
```suggestion
      .then(result => setSuccessNotification(!!result.data))
```

src/common/notification/NotificationComponent.module.css
```diff
@@ -9,5 +9,14 @@
     margin: 0;
 }
 
+/*
+    Use !important for now to shrink notification button size
+    and prevent button appearing on top of the title text.
+    Replace after DismissableNotification accepts className as prop.
+*/
+.notification button {
+    width: 50px !important;
```
ahie: > 50px

Should this be refactored into an existing(?) theme constants file?

src/common/test/testUtils.tsx
```diff
@@ -9,3 +10,11 @@ export const mountWithProvider = (children: ReactElement) =>
 
 export const shallowWithProvider = (children: ReactElement) =>
   shallow(<Provider store={store}>{children}</Provider>);
+
+/* eslint-disable  @typescript-eslint/no-explicit-any */
+export const updateWrapper = async (wrapper: any) => {
```
ahie: > updateWrapper

AFAIK this is used to test components that fetch data with apollo. Could use a library like [`wait-for-expect`](https://github.com/TheBrainFamily/wait-for-expect) instead, as is suggested in the official [Apollo docs](https://www.apollographql.com/docs/react/development-testing/testing/)?

src/pages/membership/components/membershipInformation/MembershipInformation.tsx
```diff
@@ -11,41 +11,49 @@ import NotificationComponent from '../../../../common/notification/NotificationC
...
+    }
+  );
   const [renewMembership] = useMutation<
     RenewMyYouthProfileData,
     RenewMyYouthProfileVariables
-  >(RENEW_MEMBERSHIP);
+  >(RENEW_MEMBERSHIP, { refetchQueries: ['HasYouthProfile'] });
 
-  const validUntil = convertDateToLocale(props.expirationDate);
+  const validUntil = convertDateToLocale(data?.youthProfile?.expiration);
```
ahie: > data?.youthProfile?.expiration

Possibly undefined? Will this cause unexpected behavior?

tuovinensanttu: If value is undefined function will return empty string.

src/common/test/testUtils.tsx
tuovinensanttu: That actually was first thing that I tried, but I ended up going with this solution instead. Everything I tried with `wait-for-expect` ended up in an error `When testing, code that causes React state updates should be wrapped into act(...):`. Maybe I give this package a new go when I have more time to play around with it.

PR #59, opened 2020-03-24
OM - 573 | Create youth profile improvements

tuovinensanttu:
This PR contains few bug fixes and improvements.

- Create youth profile form is prefilled with existing profile data.
- Added field for Profile language. 
- Duplicate ID error was fixed by using update mutation if user already has a profile.
- Under 15 registration was fixed. Photo permission data is not added if user is under 15 years old.


src/common/select/Select.tsx
```diff
@@ -0,0 +1,48 @@
...
+  // Without this defaultValue is not set / language wont get updated
+  const value = props.options.find(option => option.value === props.value);
+
+  return (
+    <div className={classNames('wrapper', props.className)}>
+      <label htmlFor={props.name} className="select-label">
+        {props.labelText}
+      </label>
+
+      <ReactSelect
```
Jokinen: I am not knowledgeable of the design in more depth, but it's currently a filter box while it only has three options. Setting it up in this way makes it more vulnerable for accessibility quirks because the component has to convey a lot more information to screen reader users: current search, search options list, currently selected option etc. With Varaamo, the combination of the extra information broke this component for some screen readers--some information was not communicated expectedly and some was repeated.

You could consider whether it'd be better to just use a regular HTML select here as the current usecase for this select has a very limited set of options

src/i18n/en.json
```diff
@@ -70,7 +70,9 @@
     "addInfo": "Additional info",
     "schoolName": "School name",
     "schoolClass": "School name",
+    "language": "ENGLISH",
```
Jokinen: This doesn't seem like the right place to store this information. I am assuming that you are using this translation to set the user's language somewhere.

src/pages/membership/components/createYouthProfile/CreateYouthProfile.tsx
```diff
@@ -107,47 +86,76 @@ function CreateYouthProflle({ tunnistamoUser }: Props) {
...
+        .catch((error: Error) => {
+          Sentry.captureException(error);
+          setShowNotification(true);
+        });
+    }
   };
+  // These allow us to set initial value of languageAtHome & profileLanguage
+  // to users current language.
+  const currentLangForProfile: Language = t('registration.language');
+  const currentLangForYouth: YouthLanguage = t('registration.language');
```
Jokinen: The `useTranslation` hook returns the `i18n` instance that contains information about current language/locale. You can use this information to get the current language in the format you would like.

Pseudocode:
```js
function formatLocal(locale) {
  switch(locale) {
    case 'fi':
      return 'FINNISH';
    default:
      return 'FINNISH';
}
```

src/pages/membership/components/createYouthProfileForm/CreateYouthProfileForm.tsx
```diff
@@ -116,42 +118,15 @@ function CreateYouthProfileForm(props: Props) {
...
-          approverFirstName: values.approverFirstName,
-          approverLastName: values.approverLastName,
-          approverPhone: values.approverPhone,
-          approverEmail: values.approverEmail,
-          languageAtHome: values.languageAtHome,
-          photoUsageApproved: values.photoUsageApproved,
-        });
-      }}
+      onSubmit={(values: FormValues) =>
+        props.onValues({ ...values, birthDate: props.profile.birthDate })
```
Jokinen: This is fine by me, but I think that you should not have the `birthDate` value within the form state at all. You are currently using the form's model as a kind of view layer because you are using it to display a human readable date. When finally sending the values, you have to overwrite the data.

You could remove it entirely from `Formik` and just have the input reflect the value it receives from the API directly.

tuovinensanttu: Removed `birthDate` from form data and passed value directly to TextField.

src/pages/membership/components/createYouthProfile/CreateYouthProfile.tsx
tuovinensanttu: I removed languages from translation files and created a helper for getting right value.

PR #67, opened 2020-04-22
OM-745 |  Add modify functionality

tuovinensanttu:
User can now edit & save their information. I also fixed OM-714, if user has existing profile, selected language is transferred to registration form. 

Now that `YouthProfileForm` has multiple use cases I added few tests to check functionality. 

Added new tunnistamo URL as well.


src/pages/membership/components/editYouthProfile/EditYouthProfile.tsx
```diff
@@ -0,0 +1,125 @@
...
+          },
+        },
+      },
+    };
+
+    updateProfile({ variables })
+      .then(() => {
+        history.push('/membership-details');
+      })
+      .catch((error: Error) => setShowNotification(true));
```
badrange, line 83: Does setShowNotification push errors to sentry? If not, did you consider pushing the errors to sentry - it might help us catch errors caused by bugs in the API for example.

tuovinensanttu: Oh damn, that was my intention but seems that I forgot to do it...

PR #68, opened 2020-04-24
OM-535 | SEO optimization.

tuovinensanttu:
Most of the texts are just placeholders for now & all the images and twitter related stuff will be added later. I added `Disallow: /` to robots.txt, which should block search engine access for now.

### DIFF
--- a/src/common/layout/PageLayout.tsx
+++ b/src/common/layout/PageLayout.tsx
@@ -3,7 +3,9 @@ import classNames from 'classnames';
 
 import Header from '../header/Header';
 import HostingBox from '../hostingBox/HostingBox';
+import { useTranslation } from 'react-i18next';
 import Footer from '../footer/Footer';
+import { HelmetProvider, Helmet } from 'react-helmet-async';
 import styles from './PageLayout.module.css';
 
 type Props = {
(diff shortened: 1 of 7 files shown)


badrange:
This looks good - in Finnish

Did you consider other languages or is that out of scope for this issue?

tuovinensanttu:
> This looks good - in Finnish
> 
> Did you consider other languages or is that out of scope for this issue?

Oh yeah, that totally skipped my mind! I'll add those as well!

Jokinen:
Interesting comment from @badrange 

Are social media parsers generally able to go through JS code? This template is static and for non-js crawlers will only ever this content, right 🤔

But maybe it's whatever--Google can crawl the description even if it is set from `react-helmet`.

badrange:
I haven't found a way to make social media sharing work, so I've just put Finnish into index.html since most of our users are Finnish speakers. 

But Google SERP works nicely with react-helmet!

badrange:
So if the scope of this issue is sharing, it is now done, if the scope is looking like "something" on google search result pages, react-helmet will help

tuovinensanttu:
I think adding react-helmet is in the scope of this issue and needs to be done sooner or later. I'll continue working on this issue.

src/common/layout/PageLayout.tsx
```diff
@@ -12,19 +14,28 @@ type Props = {
...
-          : styles.adultBackground
-      )}
-    >
-      <Header />
-      <HostingBox className={styles.hostingBox}>{props.children}</HostingBox>
-      <Footer />
-    </div>
+    <HelmetProvider>
+      <Helmet>
+        <html lang={i18n.languages[0]} />
```
Jokinen, line 22: Good job on including this 👏

src/common/layout/PageLayout.tsx
```diff
@@ -12,19 +14,28 @@ type Props = {
...
+        <title>{t('appName')}</title>
+        <meta name="description" content={t('login.helpText')} />
```
Jokinen, line 24: I expected to see open graph tags here as well 🤔

tuovinensanttu: Good point, added tags to this file as well.

PR #69, opened 2020-04-27
OM-528 | Update README

tuovinensanttu:
- Updated links
- Added instructions for contributing.
- Changed scripts instructions to use yarn instead of npm.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,11 @@
+Contributions are accepted as pull requests.  Please observe our coding
+practices at https://github.com/City-of-Helsinki/bestpractice/ .
+Please make your pull requests short, elegant and only handling one
+issue at a time!
+
+<!-- TODO: Uncomment below after we have service email address -->
+<!-- If you make a pull request, you may also want to contact
+<INSERT SERVICE'S ROLE ADDRESS> to tell about your contribution. -->
+ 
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
+Contributions are accepted as pull requests.  Please observe our coding
```
Jokinen, line 1: There's a double space here. I was a graphic designer for a while and I had to lay out a book. The author had left in double spaces everywhere which I had to fix by hand. So what I am saying is that it doesn't really matter, I just don't like double spaces anymore :D

CONTRIBUTING.md
```diff
@@ -0,0 +1,11 @@
...
+practices at https://github.com/City-of-Helsinki/bestpractice/ .
+Please make your pull requests short, elegant and only handling one
+issue at a time!
```
Jokinen, line 4: >elegant and only handling one issue at a time!

You probably copy pasted this from somewhere, but I think that using "handling" here is not good English 🤔Sounds a bit like it could be Finglish.

>Please make your pull requests short, elegant and only hande one issue at a time!

tuovinensanttu: Yup, it's template mentioned in confluence. I made few improvements based on your suggestions.

PR #70, opened 2020-04-27
OM-536 | Add analytics

tuovinensanttu:
Added Matomo to project. I added "simplified" titles for few pages, some of the existing ones weren't suited for analytics use.


src/common/layout/PageLayout.tsx
```diff
@@ -10,11 +11,23 @@ import styles from './PageLayout.module.css';
...
 
-function PageLayout(props: Props) {
+function PageLayout({ children, title = 'appName', background }: Props) {
   const { t, i18n } = useTranslation();
+  const { trackPageView } = useMatomo();
+
+  const pageTitle =
+    title !== 'appName' ? `${t(title)} - ${t('appName')}` : t('appName');
+
+  useEffect(() => {
```
Jokinen, line 25: This effect does not have a deps array, so it's run on every render. By adding an empty deps array it is ran on mount. By adding `pageTitle` and `href` it'd be run when those two values change.

I don't know much about matomo so I don't know if this is an issue.

src/App.tsx
```diff
@@ -25,38 +26,52 @@ loadUser(store, userManager).then(async user => {
...
 
+const instance = createInstance({
+  urlBase: 'https://analytics.hel.ninja/',
+  siteId: 59,
+});
+
+// Prevent non-production data from being submitted to Matomo
+// by pretending to require consent to process analytics data and never ask for it.
+// https://developer.matomo.org/guides/tracking-javascript-guide#step-1-require-consent
+if (process.env.NODE_ENV !== 'production') {
```
badrange, line 37: With this solution you'll be sending data from the review & staging environments to matomo. That's ok until we go into production.. especially when you want to test and verify that it works. 

I would recommend adding a new REACT_APP environment variable, for example REACT_APP_ENVIRONMENT where you only submit data for production.

src/common/layout/PageLayout.tsx
tuovinensanttu: I somehow managed to ignore run time warning by IDE threw about this. Dependency array is now added.

src/App.tsx
tuovinensanttu: Good points! I added event tracking to registration as well. I also added REACT_APP environment variable.

PR #71, opened 2020-04-28
OM 745 | Youth can remove approver information

tuovinensanttu:
Fixed issue where under 18 years old could change / remove approver information while editing profile. Now that age is determined by value coming from props, I added few tests based on user's age.

### DIFF
--- a/src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
+++ b/src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
@@ -41,7 +41,14 @@ const schema = Yup.object().shape({
   approverFirstName: Yup.string()
     .min(2, 'validation.tooShort')
     .max(255, 'validation.tooLong')
-    .when(['birthDate'], (value: string, schema: Yup.StringSchema) => {
+    .when(['birthDate'], (value: string, schema: Yup.StringSchema) => return isConsentRequired(birthDate, schema)
+    
+    const isConsentRequired = (birthDate:string, schema: Yup.StringSchema) => {
+      const userAge = differenceInYears(new Date(), new Date(value));
+      return userAge < ageConstants.ADULT
+        ? schema.required('validation.required')
+        : schema;
+    }
       const userAge = differenceInYears(new Date(), new Date(value));
       return userAge < ageConstants.ADULT
         ? schema.required('validation.required')
--- a/src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
+++ b/src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
@@ -12,6 +12,13 @@ import { Language, YouthLanguage } from '../../../../graphql/generatedTypes';
 import styles from './YouthProfileForm.module.css';
 import Button from '../../../../common/button/Button';
 
+const isConsentRequired = (birthDate: string, schema: Yup.StringSchema) => {
+  const userAge = differenceInYears(new Date(), new Date(birthDate));
+  return userAge < ageConstants.ADULT
+    ? schema.required('validation.required')
+    : schema;
+};
+
 const schema = Yup.object().shape({
   firstName: Yup.string()
     .min(2, 'validation.tooShort')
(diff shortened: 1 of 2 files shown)


src/pages/membership/components/youthProfileForm/__tests__/YouthProfileForm.test.tsx
```diff
@@ -1,29 +1,33 @@
...
+    profileLanguage: Language.FINNISH,
+    languageAtHome: YouthLanguage.FINNISH,
+    schoolName: 'School',
+    schoolClass: '',
+    photoUsageApproved: 'false',
+    approverEmail: 'test@test.fi',
+    approverLastName: 'TestApprover',
+    approverFirstName: 'TestFirstName',
+    approverPhone: '0501234567',
+    ...values,
```
badrange, line 29: Nice trick!

src/pages/membership/components/youthProfileForm/__tests__/YouthProfileForm.test.tsx
```diff
@@ -69,3 +73,37 @@ test('Profile is used for profile editing', async () => {
   expect(buttonRow.children().length).toEqual(2);
   expect(terms.length).toEqual(0);
 });
+
+describe('Form fields & texts based on user age', () => {
+  test('user age is >= 13 < 15', () => {
```
badrange, line 78: These tests are going to save us so many times in the future

src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
```diff
@@ -45,30 +41,34 @@ const schema = Yup.object().shape({
   approverFirstName: Yup.string()
     .min(2, 'validation.tooShort')
     .max(255, 'validation.tooLong')
-    .when([], (schema: Yup.StringSchema) => {
-      return AGE < ageConstants.ADULT
+    .when(['birthDate'], (value: string, schema: Yup.StringSchema) => {
```
badrange, line 44: You have the same check (if I'm reading correctly) copied to three places. Would it make sense to split that out to a separate boolean, so we could have something like `when(isConsentRequired)` ?

tuovinensanttu: Because schema is outside of the component, saving value to boolean would require adding it to form values, which on it's own way doesn't sound like a good idea. 

I could change it to something like this:
```suggestion
    .when(['birthDate'], (value: string, schema: Yup.StringSchema) => return isConsentRequired(birthDate)
    
    const isConsentRequired = (birthDate:string) => {
      const userAge = differenceInYears(new Date(), new Date(value));
      return userAge < ageConstants.ADULT
        ? schema.required('validation.required')
        : schema;
    }
```
Creating a custom validation method would be other option, but doing that seems like a hassle with TypeScript i'm not sure if it's possible to access other fields value ([example](https://github.com/jquense/yup/issues/312#issuecomment-442854307))

badrange: This was exactly what I meant! 

In case the rules for consent changes we only need to update the logic in one single place.

src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
```diff
@@ -45,30 +41,41 @@ const schema = Yup.object().shape({
   approverFirstName: Yup.string()
     .min(2, 'validation.tooShort')
     .max(255, 'validation.tooLong')
-    .when([], (schema: Yup.StringSchema) => {
-      return AGE < ageConstants.ADULT
+    .when(['birthDate'], (value: string, schema: Yup.StringSchema) => return isConsentRequired(birthDate, schema)
```
badrange, line 44: Hm, this last commit doesn't really help does it. What you're saying is that we can't share the isConsentRequired logic between the different fields in any simple way due to typing issues? 

Then I think we should revert my last commit and merge it as it was before..

tuovinensanttu: I think it should work, I'll fix it manually.

tuovinensanttu: The helper function is now inside the .when() and propably crashed everything.

tuovinensanttu: Seems to be working now 👍

badrange: Great!

PR #73, opened 2020-05-06
OM-542 | Add update-translations script

Jokinen:
Adds `yarn update-translations` script from [helsinki-utils-js](https://github.com/City-of-Helsinki/helsinki-utils-js). It creates new translation message files based on the current version of translation source and overwrites existing files with them.

I haven't fetched the latest translations in this PR--should I?

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,7 +32,6 @@
     "graphql.macro": "^1.4.2",
     "hds-core": "^0.4.1",
     "hds-react": "^0.5.4",
-    "helsinki-utils": "https://github.com/City-of-Helsinki/helsinki-utils-js.git#0.1.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -32,6 +32,7 @@
     "graphql.macro": "^1.4.2",
     "hds-core": "^0.4.1",
     "hds-react": "^0.5.4",
+    "helsinki-utils": "https://github.com/City-of-Helsinki/helsinki-utils-js.git#0.1.0",
```
badrange, line 35: Wonder if this can be added to devDependencies? Doesn't make a big difference, but I notice that it feels "safer" to update those..

Jokinen: You are right 👍

PR #75, opened 2020-05-11
YM-163 | Add country code

tuovinensanttu:
Added country code to details view and forms.
Some of the Apollo generated types were changed. I needed to refactor few components to get everything working again.


src/pages/membership/components/approveYouthProfile/ApproveYouthProfile.tsx
```diff
@@ -92,7 +92,7 @@ function ApproveYouthProfile(props: Props) {
                 data?.youthProfileByApprovalToken?.profile?.firstName || '',
               lastName:
                 data?.youthProfileByApprovalToken?.profile?.lastName || '',
-              address: getAddress(data),
+              address: getAddress(data, i18n.languages[0]),
```
Jokinen, line 95: I've used `i18n.language` to get the current language. For some cases it doesn't work. I am wondering if this is such a case? I am asking purely out of interest, because I do not understand how they differ :D

tuovinensanttu: I'm not 100% sure, mostly because documentation won't explain it either. They do however mention that getting current language `i18n.languages[0]` is preferred. [Docs](https://www.i18next.com/overview/api#language)

badrange: > If you need the primary used language depending on your configuration (whilelist, load) you will prefer using i18next.languages[0].

So in this case, we always want to default to fi since we are in Finland. Nice find in the docs - the other option could have been some configuration (always default to fi) or hard coding.

PR #79, opened 2020-05-14
YM-198 | Language selection changes.

tuovinensanttu:
Changed language order, current language is not shown.

### DIFF
--- a/src/i18n/languageSwitcher/LanguageSwitcher.tsx
+++ b/src/i18n/languageSwitcher/LanguageSwitcher.tsx
@@ -6,8 +6,8 @@ import styles from './LanguageSwitcher.module.css';
 
 const languages = [
   { code: 'fi', label: 'Suomi' },
-  { code: 'en', label: 'English' },
   { code: 'sv', label: 'Svenska' },
+  { code: 'en', label: 'English' },
 ];
 
 type Props = {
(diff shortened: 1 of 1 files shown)


src/i18n/languageSwitcher/LanguageSwitcher.tsx
```diff
@@ -26,15 +26,19 @@ function LanguageSwitcher(props: Props) {
...
-          <button
-            type="button"
-            lang={lang.code}
-            onClick={() => setLanguage(lang.code)}
-          >
-            {lang.label}
-          </button>
-        </li>
+        <React.Fragment key={lang.code}>
+          {lang.code !== i18n.languages[0] ? (
```
Jokinen, line 30: Have you considered filtering the unwanted language? It could make for an easier to understand component.

**Also**

Your language code as in the form `language_code` whereas the `i18n.languages` array may be in the format `language_code-locale`.

tuovinensanttu: Yep, you were right. If there were no cookies, localStorage or navigation parameters present, `i18next-browser-languagedetector` defaulted language format to `fi-FI`. I added these values to all languages and changed ternary check to filter method.

PR #82, opened 2020-05-15
YM-196 | Refactor date input

Jokinen:
- Fix non-numeric characters being allowed on iOS
- Fix negative numbers being allowed on any platform
- Set maximum length for numbers in character counts. Two for date and
  month inputs, four for year input.
- Automatically change focus into next input when previous input has
  been filled

Screenshot 2020-05-18 at 8 56 46
Screenshot 2020-05-18 at 8 56 56
Screenshot 2020-05-18 at 9 06 19

----
### Caveats

- Automatic change of input only works for dates and months that are two characters long. In the case that the user has been born on day 1-9, and/or months of 1-9, this functionality will provide reduced benefit. Because we represent the input as `type="number"` leading zeroes can not be inputted and hence single digit numbers can never be "two characters" long.
- Non-sensical dates are allowed. The user can input a date like so: `40.12.2005`. By some whim of JS, this gets parsed into a date that may then validate or not based on the resulting date. I was not sure if this was allowed previously--if it was not then I'll create a fix. Validating the maximum date number in convoluted and would require pulling in a library in my opinion.
- Dates are parsed into zulu time sticking to default JS conventions. This means that an input of `1.1.2020` made in the Helsinki timezone would yield (depending on DST) `12.12.2019T21:00:00Z`. This is fine as long as the libraries used to count ages in the frontend, and then in turn the backend expect birthdate to be a "timezone" aware date. Looking at the code context, it seems like this expectation can be made.


src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,240 @@
...
+  monthInputName,
+  yearInputName,
+  dateComponent,
+  monthComponent,
+  yearComponent,
+  dateInputLabel,
+  monthInputLabel,
+  yearInputLabel,
+}: Props) {
+  const currentDate = getDateComponents(value);
```
badrange, line 144: Is currentDate the date provided by the user or today's date? Could be helpful to change the name of the variable to something like `submittedDate` `userProvidedDate` or similar.

src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,240 @@
...
+  const focusWhen = (
+    condition: boolean,
+    element: React.RefObject<HTMLInputElement>
+  ) => {
+    if (condition && element.current) {
+      element.current.focus();
+    }
+  };
+
+  const handleDateChange = (event: React.FormEvent<HTMLInputElement>) => {
```
badrange, line 179: Is this handling the whole date like 2020-12-25 OR the day of month? Maybe you can rename to handleDayChange in that case.

src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,240 @@
...
+  yearInputName,
+  dateComponent,
+  monthComponent,
+  yearComponent,
+  dateInputLabel,
+  monthInputLabel,
+  yearInputLabel,
+}: Props) {
+  const currentDate = getDateComponents(value);
+  const [cachedDate, setCachedDate] = React.useState<DateObject>(currentDate);
```
badrange, line 145: is the date cached in the browser or is it only used in the state? If only state, I think another name could be more clear, but I can't think of an improvement so cached is fine

src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,240 @@
...
+    updateDate(null, null, nextYear);
+  };
+
+  const formattedMonth =
+    cachedDate.month !== undefined
+      ? (cachedDate.month + 1).toString()
+      : undefined;
+
+  return React.createElement(wrapper, { isInvalid }, [
+    React.createElement(dateComponent, {
```
badrange, line 205: If this component is actually about the day of the month, I think dayComponent would be a better name, so that it is easier to separate it from a "full/normal date"

src/common/dateInput/DateInputLogic.tsx
Jokinen: It's not necessarily neither. It's the currently selected date from the perspective of the state. Due to how a `Date` object can be created--you need to know the date, month and year--we need to have two representations of a date at the same time:
* The values that represents the inputs' states (cached date: A temporary saving mechanism for parts of dates, until we can build a complete date object based on them, and send that to the consumer)
* The actual date that is this component communicates to its consumers (current date: this is the ultimate truth)

In this implementation the "current" date is not used for anything else than initialisation--but it's still a rather important concept for the architecture of this component. This is because we have two states that we need to sync at the correct moments. It's more or less an antipattern, but in this case it's hard to avoid and this is mostly the approach native HTML inputs take as well--one internal value and one external value. For instance we would need to sync the `currentDate` value into the inputs' state if we used this component in a context where some other input controlled the same value as this input does.

But alas the naming here stands to be improved, I only provided the background so you can make a better judgment on an appropriate name. Maybe something along the lines of internal/external. Cached -> temporary etc.

src/common/dateInput/DateInputLogic.tsx
Jokinen: With JS Dates, date = day of month and day = day of week. I don't know how dogmatically this should be followed and where is the correct point to switch naming. My rationale here was that this input control's the date value of a JS Date, so we want to reflect that in the naming as well.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDate
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getDay

src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,246 @@
...
+  // is still able to input a negative number by hand if they choose.
+  min: 0,
+  // 1) Ensures that a number pad is shown on iOS instead of the regular
+  // keyboard
+  // 2) Blocks iOS devices from inputting characters other than numbers
+  pattern: '\\d*',
+};
+
+type DateObject = {
+  date?: number;
```
badrange, line 30: Would it be even better to use `day` or `dayofMonth`? I am quite sure `date` will cause confusion, since date is usually a specific day on a calendar: 

https://en.wikipedia.org/wiki/Calendar_date

src/common/dateInput/DateInputLogic.tsx
```diff
@@ -0,0 +1,246 @@
...
+  if (!date) {
+    return {};
+  }
+
+  const monthDate = date.getDate();
+  const month = date.getMonth();
+  const year = date.getFullYear();
+
+  return {
+    date: monthDate,
```
badrange, line 82: if `monthDate` is the day of the month, maybe we could even use `day` here?

src/common/dateInput/DateInputLogic.tsx
Jokinen: I agree that that's the case generally--but the developer interacting with this code should be aware of how `Date` object in JS behave, because this component is in essence a control for creating JS `Date`s. Because of that, I'd value the development context here more than the "human" (:D) context. I.e. the outward facing naming should adopt the general pattern, but the internal naming should align with the concepts that are important for the internal implementation. In that sense I think the implementation should be looked from this perspective: _I am a developer of this component_, instead of this perspective _I am a human_/_I am a consumer of this component_.

But alas it's not a big enough distinction that I'd want to put my foot down, so I'll change it.

Jokinen:
We chatted about the other issue @badrange found through Slack. It's a problem that's caused by

>Automatically change focus into next input when previous input has
been filled

In essence the problem scenario is this: a user using the arrow keys can't conveniently set the value of a number input to a value greater than 10, because the focus will be shifted to the next input for each two digit number they have to "go through" to arrive at their target number.

There are a few routes we could use to mitigate this issue. We can re-implement the arrow key listeners and provide our custom spinner controls for the component. We can try to look into if the event object has some consistent and reliable piece of information that allows us to deduce the source of the event. We can also think about using a text type input instead of numbers which would remove these controls--this would allow leading zeroes as well (currently the error message instructs the user to input `dd.mm.yyyy`, but if your birthday is a one digit number, you won't be able to abide by this instruction, strictly speaking).

src/common/dateInput/DateInputLogic.tsx
badrange: Oh, thank you for this input, I should have known this!

PR #85, opened 2020-05-18
YM-208 | Remove ToS page

Jokinen:
Removes ToS page from project and pulls new translations that are needed for ignoring it.


src/pages/membership/components/approveYouthProfileForm/ApproveYouthProfileForm.tsx
```diff
@@ -208,13 +207,13 @@ function ApproveYouthProfileForm(props: Props) {
...
+                    <a
+                      href={t('registry.descriptionLink')}
                       target="_blank"
                       rel="noopener noreferrer"
                     >
                       {t('approval.approveTermsText_link')}
-                    </Link>
+                    </a>
                     {t('approval.approveTermsText_2')}
                   </span>
```
tuovinensanttu, line 218: These translation keys still point to the old terms text. I think you could replace these with `Trans` component + `registration.approveTerms`  key.

Jokinen: Good point. Hmn, I'll check with whether we can get he new translations into the Google sheet in the same format as the other link.

PR #89, opened 2020-05-26
YM-226 | Tests for ApproveYouthProfileForm.

tuovinensanttu:
Added few tests for `ApproveYouthProfileForm` component.


Jokinen:
You forgot to include the new snapshot in these changes, which causes the CI to fail.

tuovinensanttu:
> These test look good! As a suggestion, here are some other cases I could see as good candidates for a first version of the tests:
> 
> * Behaviour of photo usage approval field
> * One happy path tests where the form submits successfully with some set of valid data

I'll add submit test later. At the moment there is something weird going on with formik + test-library combo and I can't get rid of errors.

PR #90, opened 2020-05-26
YM-227 | PhotoUsageApproved is now passed to backend.

tuovinensanttu:
- When editing profile, `photoUsageApproved` field is sent to backend.
- When adding missing field I noticed update mutation was missing Sentry capturing so I added that as well.


tuovinensanttu:
> Not entirely sure how the code works here in more detail so I am not sure how easy it would be, but as you have been adding tests to other views, maybe you could add a test that guards against this fix breaking again?

I could change `handleOnValues` to some kind of helper that forms correct input object. That way it would be possible to test that all the fields are present and in correct places.

Jokinen:
Nice 👍

PR #91, opened 2020-05-26
OM-200 | Logo is now a link

tuovinensanttu:
- Helsinki logo is now a link that points to home page.
- Added margin between swedish logo & service name

### DIFF
--- a/src/common/helsinkiLogo/HelsinkiLogo.tsx
+++ b/src/common/helsinkiLogo/HelsinkiLogo.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
+import { useHistory } from 'react-router';
 import classNames from 'classnames';
 
 import styles from './HelsinkiLogo.module.css';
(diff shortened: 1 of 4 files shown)


src/common/helsinkiLogo/HelsinkiLogo.tsx
```diff
@@ -9,16 +10,23 @@ type Props = {
 };
 function HelsinkiLogo({ className }: Props) {
   const { t, i18n } = useTranslation();
+  const history = useHistory();
+
+  const navigateToHomePage = () => history.push('/');
 
   if (i18n.languages[0] === 'sv')
```
Jokinen, line 17: Not in this PR specifically, but might this benefit from the helper?

src/common/helsinkiLogo/HelsinkiLogo.tsx
```diff
@@ -9,16 +10,23 @@ type Props = {
...
     return (
       <span
+        role="button"
```
Jokinen, line 20: If this element does what a link does, why not have it as an anchor? Is it not semantic?

tuovinensanttu: Took another look at the component today and decided to replace spans with `Link` components. I'll just double check before merging that footer is supposed to be link as well.

src/common/helsinkiLogo/__tests__/HelsinkiLogo.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+};
+
+describe('renders correct logo based on language', () => {
+  test('finnish logo', () => {
+    const wrapper = getWrapper();
+    expect(wrapper.find('.logoFi').length).toEqual(1);
+  });
+
+  test('swedish logo', () => {
+    i18n.changeLanguage('sv');
```
Jokinen, line 22: How well are these test sandboxed, I can't remember 🤔I think that once you change the language, it'll remain as the language you have switched to for the ensuing tests. For consistency, maybe it could be worth it to set the language back into the default language with an `afterEach` etc.

tuovinensanttu: Next test below this one has language set to `fi`, so it seems that language change is affecting only single test.

PR #99, opened 2020-06-03
YM-222 | User session

tuovinensanttu:
- If multiple tabs are open, user actions (login / logout) will affect all tabs.
- While doing this I ran into bug, where `birthDate` cookie would have false value e.g `;  _randomValue`. Removing `.` from regex improved matching and fixed the issue. It's really small fix so I though I would apply it here.


src/App.tsx
```diff
@@ -49,6 +52,22 @@ if (process.env.REACT_APP_ENVIRONMENT !== 'production') {
 type Props = {};
 
 function App(props: Props) {
+  window.addEventListener('storage', event => {
```
Jokinen, line 55: Should this be in an `useEffect`? That way you could remove the event listener on dismount.

tuovinensanttu: When it comes to `localStorage` it seems that value changes won't trigger, not even inside `useEffect`.  This solution was only way that I managed to achieve automatic functionality.

PR #116, opened 2020-06-25
YM-84 | Multiple addresses

tuovinensanttu:
Added multiple address support. User can add, update and remove additional addresses. The `Add another address` button is not yet placed correctly. There is a new [design](https://futuricedesigners.invisionapp.com/share/VXU5VIHQDGS#/screens/420622793) for this page, but it because this form is nested very deeply, it requires a bit more work and I would like keep this PR somewhat under control.


src/pages/membership/components/youthProfileForm/YouthProfileForm.tsx
```diff
@@ -214,34 +302,147 @@ function YouthProfileForm(componentProps: Props) {
...
+                        addressType: AddressType.OTHER,
+                        __typeName: 'AddressNode',
+                      })
+                    }
+                  >
+                    {t('registration.addAddress')}
+                  </Button>
+                </React.Fragment>
+              )}
+            />
```
badrange, line 445: This file is pretty big now. Wonder if there is a convenient way to split it up?

tuovinensanttu: When the layout is updated and information is divided to multiple "boxes",  this gives nice opportunity to split it up.

badrange: Sounds like a good plan.

PR #118, opened 2020-06-26
YM-243 | Validate youth profile form fields on blur

Jokinen:
This way we avoid situations where the form does not send due to an
error, but the error is out of screen for the user, causing confusion.

----

As far as I can tell, there were no other inputs than `TextInput`s which could end up in an erroneous state.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,10 @@
+# Changelog
+All notable changes to this project will be documented in this file.
+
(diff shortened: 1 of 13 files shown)


Jokinen:
I had forgotten an `errorMessage` field in the props of `FormikTextInput`. In the first iteration I used the prop for passing the error message, but later decided to bake it into the component.

PR #119, opened 2020-06-29
YM-243 | Fix fields not being validated on blur

Jokinen:
The phone number and all the approver fields were not validated on
blur, because they had not been transitioned to use FormikTextInput.


Jokinen:
Travis wouldn't run all checks on this PR.

PR #121, opened 2020-06-30
YM-265 | Refactor app structure towards best practises

Jokinen:
In order to implement YM-265, I have to alter the page structure in order to be able to offer more flexibility for the layer in which a white background colour is applied. While trying different approaches I found it as difficult to arrive at an approach that was clear about responsibilities. I think that this was in part because I had difficulties combining the current application structure and the [goal](https://dev.hel.fi/coding-standards#folder-structure).

I don't want to refactor this part of the app in a way which will need work after we transition to the best practice approach. But I also don't want to introduce "another style" into this application. Based on my experiences with Varaamo, you don't want to leave a refactor halfway if you can help it--and if you must, you should create documentation to facilitate that.

So after all that I tested how far I could get within an hour--the refactor didn't end up being that major. So with this PR I am suggesting that we do this refactor now and get it out of the way.


badrange:
I have to admit that I did not look extremely carefully at this, but I did test it locally and it does look good to me.

PR #123, opened 2020-07-01
YM-265 | Divide youth profile form into sections

Jokinen:
Previously some routes were handled within a domain and were wrapped with custom logic. My task was to create a style change that required the functionality of the application to stay the same, yet requiring a change in how different app layers were composed.

In order to complete the fix, I refactored the pages in question from a more inheritance relying pattern into a more composable one.

You can use the GitHub GUI to browse changes commit by commit.

1. I flattened the route structure in the root of `App`. This way I was able to compose different styles for the form views as compared to the other views that did not require sections. This way all the routes also became easier to rediscover. This approach also discourages entangling views together in a way which is not composable.
     1. In order to facilitate some of the business logic for `youthProfile` pages, I created a new route component `AppYouthProfileRoute` which checks if the user has a profile and redirects them accordingly if they do not.
1. `PageLayout` is no longer declared multiple times, but is declared once for the whole application. This allows navigation and footer elements to always stay visible when the user has loaded the application and begins navigation between pages. This change became timely due to the change 1. This should decrease the chance for visual flickers which make the application seem slower.
      1. `PageLayout` is refactored to use a CSS grid layout. I did this in order to facilitate the applying of a `background` from lower within the tree, which was likely one reason for declaring `PageLayout` multiple times. This also means that we do not have to define for instance max-width separately for header, footer and so on. I moved `header` and `footer` within `common/layout` and made them aware of the main grid.
     1. I added a `LoadingPage` component which has sensible defaults and knows how to hook on to the main grid. It's deferred so that it's content won't be loaded for very short loads, decreasing the amount of flickering in the UI which may make it feel slower.
     3. I created a new component `PageContent` which pages (or maybe the more common convention is containers) can use to define their contents. This component accepts the title prop which is used for tracking, it places the content correctly on the grid and offers a straight forward way of applying a loading indicator. Previously the `PageLayout` component marked a new page, now this component more or less the same thing.
1. Finally, I divided `YouthProfileForm` into sections after I was now easily able to toggle off the graphical elements that made it difficult to do this change elegantly.
    1. I added a `YouthProfileFormSection` component. I tried the component from HDS, but it was very different to the design file, which is why I ended up creating a custom one.
    1. I separated section's fields into their own files. This shortens the length of the original form file, which may make it easier to read. Architecturally I don't think that this approach is very good--it can happen that some of the code that's now in separate files needs to be edited together later on. This is fixing the symptom more than the cause, but some symptoms are worth alleviating I guess.
    1. I applied some general style changes to text that were on the design files.


src/domain/youthProfile/form/youthProfileFormSection.module.css
```diff
@@ -0,0 +1,12 @@
+.youthProfileFormSection {
+  --youth-profile-form-section-max-width: 1024px;
+
+  max-width: var(--youth-profile-form-section-max-width);
+  margin: 0 auto;
+  padding: var(--spacing-layout-s) var(--spacing-3-xl);
+
```
tuovinensanttu, line 7: `var(--spacing-3-xl)` this seems pretty huge on small screens.

tuovinensanttu:
Also seems that around `width: 374px` the "Fullscreen navigation" breaks a little.


Screenshot 2020-07-02 at 12 22 14

Jokinen:
Good points fixing 👍

badrange:
Nice refactoring work! This will be welcome in the future!

Jokinen:
As far as I could tell, the reason for the glitch was the large amount of padding. This caused the content of the page to overflow the `100vw` limit set for the full screen navigation. By decreasing the amount of padding, the content of the page fit inside the viewport and the menu behaved as expected.

I also changed the `width` rule of the `HostingBox` for `PageLayout` into `max-width` as the content of the page would not always have `800px` worth of space to expand into, causing overflow.

PR #126, opened 2020-07-23
YM-267 | Scroll to first error

tuovinensanttu:
- Submit button is always clickable (removed disabled status if terms was not accepted)
- Added error text form terms
- If error occurs, form will be autoscrolled to first invalid element.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,6 +21,7 @@
     "@types/react-router-dom": "^5.1.0",
     "@types/react-select": "^3.0.11",
     "@types/redux-thunk": "^2.1.0",
+    "@types/scroll-to-element": "^2.0.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -51,6 +52,7 @@
     "react-scripts": "^3.3.0",
     "react-select": "^3.0.8",
     "redux-oidc": "^3.1.5",
+    "scroll-to-element": "^2.0.3",
```
Jokinen, line 55: the first thing that comes to my mind when I see this dependency is that it looks a lot like [`Element.scrollIntoView()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scrollIntoView). Is this dep used for polyfilling?

PR #127, opened 2020-07-23
Update translations

tuovinensanttu:


Jokinen:
Updated snaphshots in order to resolve broken tests.

PR #129, opened 2020-08-10
YM-178 | Allow guardians to check additional addresses

Jokinen:


src/domain/membership/helpers/getAddresses.ts
```diff
@@ -0,0 +1,59 @@
...
+  data: YouthProfileByApprovalToken | MembershipDetails
+): Address[] {
+  const edges = findAddressesEdges(data) || [];
+
+  return edges
+    .map(edge => edge?.node)
+    .filter((node): node is Address => Boolean(node));
+}
+
+export default function getAddress(
```
badrange, line 46: I wonder if the name of this function is correct? If it returns an array of multiple addresses, I suppose getAddresses is more correct? 

Also perhaps a comment to the code would be helpful, so you can quickly see that it filters out the primary address

badrange: Maybe even `getAdditionalAddresses()` could be a descriptive and long name?

Jokinen: The data model uses `addresses`. I think that it could be confusing if we introduced terminology here that's not used elsewhere 🤔

PR #135, opened 2020-08-17
YM-293 | Remove robot.txt disallow parameter

tuovinensanttu:
Removed `Disallow` parameter from `robot.txt` file.


badrange:
> I couldn't find the relevant documentation for this. I was trying to ensure that this change only impacts `production`, which is probably does. You probably remember better than I?

The infra automatically adds a deny-all robots.txt to review & staging environments, so this should be a safe solution. 

Let's check that it works after we merge this, not safe to assume things..

PR #140, opened 2020-08-19
Use same presentation logic for array fields

Jokinen:
In this PR I add a common component for array fields. When we have a single field, we can ensure that the array fields behave and look roughly the same.

I also refactored the styles to make use of grids. That way we have less markup and less complex markup. These changes also made it easier to make the array fields consistent with the other fields. Further tweaks to the spacing should now be easier to do.


tuovinensanttu (approved):
Looks good to me 👌

PR #141, opened 2020-08-19
Use same component in details and approval view

Jokinen:
This way fixing one fixes the other and the views don't fall out of
sync when it comes ot basic details.


tuovinensanttu (approved):
Gave a little test to UI and everything is working. Code is looking solid as well. Really good job with these refactoring's, keep it up 👍

PR #142, opened 2020-08-19
OM-295 | Add og-image-url as environment variable

tuovinensanttu:


klempine (changes requested):
LGTM, but you could make the environment variable's name a bit more generic so if it's needed again somewhere else, then it will make more sense. So instead of `OG_URL`, how about `BASE_URL`?

tuovinensanttu:
> LGTM, but you could make the environment variable's name a bit more generic so if it's needed again somewhere else, then it will make more sense. So instead of `OG_URL`, how about `BASE_URL`?

True, that makes more sense. Renamed.

klempine (approved):
🥇

PR #144, opened 2020-08-24
YM-313, YM-307 | Change registration max age limit to 29

klempine:
This PR is mostly about changing the max age of the service to 29 via configuration (YM-313), but I had to update the translations to correct the validation error messages, so the app now contains the latest Youth services text fixes as well (YM-307).


tuovinensanttu (commented):
Looking good. Few tests are crashing though.

klempine:
How silly of me, I didn't realize the text changes affect the tests. Should be OK now!

tuovinensanttu (approved):
👌

PR #149, opened 2020-08-26
YM-308 | Fix non existent dates silently working

Jokinen:
Previously, if the user inputted a non-existent date (for instance a leap year date or just a day that does not exist in general), it would be parsed into a real date silently due to how the `Date` object works.

Screenshot 2020-08-20 at 15 42 52

After this change, the `DateInput` component checks whether the given date actually exists. It does this by comparing if the date part (e.g. for 2020/07/15 => 15) of the created `Date` object matches with the date the user has inputted. If not, something magical has happened and the inputted date was likely incorrect.

I also made an additional tweak which allows the user to input leading zeroes to dates. Previously leading zeroes would be parsed away. After this change, they should be retained in the rendered input. This small tweaks allows the user to input the dates as described in the `label`. Further, removing the leading zeroes may create general confusion and diminishes the value of "auto changing" input fields.


src/common/components/dateInput/DateInputRow.tsx
```diff
@@ -1,18 +1,30 @@
 import React, { useRef } from 'react';
 import classNames from 'classnames';
 
-import useFocusWithinOut from './useFocusWithinOut';
 import { WrapperComponentProps } from './DateInputLogic';
 import styles from './dateInputRow.module.css';
 
 function DateInputRow({ isInvalid, onBlur, ...rest }: WrapperComponentProps) {
   const wrapperElement = useRef(null);
```
Jokinen, line 8: This can be removed.

tuovinensanttu:
The previous bug where inputting `31.02.YYYY` seems to fixed. There seems to be a bug with months though: If user inputs date ex. `31.13.2000` it will pass user forward and translate date to `31.01.2001`.

Jokinen:
I haven't taken that case into consideration, but I don't see why it'd be any less important than the that was fixed. I'll add changes which handle the month as well.

Jokinen:
Screenshot 2020-09-04 at 13 12 17
Screenshot 2020-09-04 at 13 12 03

tuovinensanttu (approved):
Nice work 👌

PR #150, opened 2020-08-27
YM-320 | Add link to Helsinki profile privacy policy

klempine:
Updated the latest translations from the Excel which now contained the link
to Helsinki profile privacy policy + updated texts for terms approval.

Changed the terms approval so that it now adds a link which points to the
Helsinki profile privacy policy so that the user is at least minimally
notified of that registry.


Jokinen (approved):
Looks good to me 👍

PR #152, opened 2020-08-31
YM-322 | Fix profile information page crash

tuovinensanttu:
Because `MembershipDetailsFragment` was missing profile ID, it caused Apollo cache to crash when fetching new information.


Jokinen (approved):
👍

PR #155, opened 2020-09-07
YM-311 | Remove cypress add testcafe

tuovinensanttu:
Removed cypress and added testcafe.


browser-tests/BirthdayForm.ts
```diff
@@ -0,0 +1,34 @@
...
+    .expect(loginSelector.errorText.innerText)
+    .eql('Invalid date');
+
+  await t
+    .selectText(loginSelector.day)
+    .typeText(loginSelector.day, '01')
+    .selectText(loginSelector.month)
+    .typeText(loginSelector.month, '13')
+    .selectText(loginSelector.year)
+    .typeText(loginSelector.year, '2010')
```
Jokinen, line 31: This chain of commands is used a lot. I wonder if it'd be worthwhile to make a helper of it?

browser-tests/BirthdayForm.ts
```diff
@@ -0,0 +1,34 @@
+import { loginSelector } from './pages/loginSelector';
+import { testURL } from './utils/settings';
+
+fixture('Open birth date form').page(testURL());
+
+test('Test invalid dates', async t => {
```
Jokinen, line 6: Testing with this kind of detail ("wrong month dates don't work", "leap day won't work" and "wrong month won't work") suggests a more unit testing like mindset. I'm not sure if they are good test candidates for `e2e` tests. For `e2e` tests I'd consider more user facing test cases, for instance "X-Y year olds can register" and ">Z year olds can't register". You already seem to have these cases covered.

Also, invalid date behaviour has been tested with unit(/integration) tests on the `DateInput` level. In that sense the behaviour is covered. Although I only tested a single non-existent date, because I considered that to cover enough ground. I would have placed these test cases as unit/integration tests in the `DateInput` files.

browser-tests/pages/loginSelector.ts
```diff
@@ -0,0 +1,11 @@
+import { Selector } from 'testcafe';
+export const loginSelector = {
+  submitButton: Selector('button[type="submit"]'),
+  errorText: Selector('span[class="hds-text-input__helper-text"]'),
+  day: Selector('input[id="ym-date-input-date-id"]'),
+  month: Selector('input[id="ym-date-input-month-id"]'),
+  year: Selector('input[id="ym-date-input-year-id"]'),
+  paperForm: Selector(
+    'a[href="https://palvelukartta.hel.fi/fi/search?q=nuorisotalo"]'
+  ),
```
Jokinen, line 10: Some of these selectors seem relatively brittle to me. I looked through some examples, for instance berth control, and they used a similar approach. I guess that something in TestCafe makes this approach palatable?

Jokinen (approved):
I think that it could be helpful, if you updated the readme with information about the `e2e` tests. In other words, how to run them locally (commands) and details about how they are ran in the CI.

browser-tests/BirthdayForm.ts
tuovinensanttu: That is true. I'll check if I manage to make helper for this.

browser-tests/pages/loginSelector.ts
tuovinensanttu: These seemed little iffy to me too. I'll replace this with `test-id` attribute.

browser-tests/BirthdayForm.ts
tuovinensanttu: I'll remove these.

PR #158, opened 2020-09-11
YM-352 | Improve authentication callback error handling

Jokinen:
Previously the error handling in the authentication callback didn't handle some known problem cases
* Device time off by more than 5min
* User declines permission

I just copied the logic over from Kukkuu.

The real issue was that the callback used `history.push` when it redirected the user into another view. This left the `/ callback` page into the browser history. We suspect that users sometimes accidentally access this page again when they use the browser's back button to navigate.

I also noticed that we were not using `main` at all, so I added it into the `<PageContent>` component.

### DIFF
--- a/src/i18n/fi.json
+++ b/src/i18n/fi.json
@@ -33,10 +33,10 @@
   },
   "authentication": {
     "deviceTimeError": {
-      "message": "Usko tai älä usko mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan."
+      "message": "Usko tai älä usko, mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan."
     },
     "permRequestDenied": {
-      "message": "You need to approve our request to access your Tunnistamo profile in order to use this service. Please try logging in again."
+      "message": "Sinun täytyy hyväksyä pyytämämme oikeudet Tunnistamossa käyttääksesi tätä palvelua. Ole hyvä ja yritä kirjautua uudelleen."
     },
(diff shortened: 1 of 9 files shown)


src/i18n/fi.json
```diff
@@ -31,6 +31,16 @@
     "schoolInfo": "Koulu ja luokka",
     "title": "Hyväksy jäsenyys"
   },
+  "authentication": {
+    "deviceTimeError": {
+      "message": "Usko tai älä usko mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan."
+    },
+    "permRequestDenied": {
+      "message": "You need to approve our request to access your Tunnistamo profile in order to use this service. Please try logging in again."
```
tuovinensanttu, line 39: This should be in finnish.

src/i18n/fi.json
```diff
@@ -31,6 +31,16 @@
     "schoolInfo": "Koulu ja luokka",
     "title": "Hyväksy jäsenyys"
   },
+  "authentication": {
+    "deviceTimeError": {
+      "message": "Usko tai älä usko mutta et voi kirjautua sisään koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan."
+    },
```
tuovinensanttu, line 37:
```suggestion
      "message": "Usko tai älä, mutta et voi kirjautua sisään, koska laitteesi kello on yli 5 minuuttia väärässä. Säädä kelloa ja kokeile uudestaan."
    },
```

tuovinensanttu (approved):
Few minor things, otherwise looking good 👌

Jokinen:
🤦‍♂️ fixing

PR #159, opened 2020-09-17
YM-347 | Add helpers text relating to additional approvers

tuovinensanttu:
Additional information text is show, if multiple guardians are added.


src/common/components/arrayFieldTemplate/ArrayFieldTemplate.tsx
```diff
@@ -24,6 +26,7 @@ function ArrayFieldTemplate({
   onPushItem,
   space = 'm',
   listSpace = 'l',
+  additionalGuardianHelperText,
```
Jokinen, line 29: I made comments in the admin-ui PR. I think that the same things apply here.

Jokinen (approved):
Looks good 👍

PR #160, opened 2020-09-22
YM - 359 | Registration e2e tests

tuovinensanttu:
In `jässäri-admin` all dotenv files were combined into single one `.env`. I made the same change here.

Added `e2e` tests for registration form. Log in credentials are the same as in `jässäri-admin`, you can contact me for those. Error testing is pretty minimal, it only checks amount of errors user action triggered (in this case submitting empty form and counting all required fields).  

Every time tests are run new profile is created. Remember to delete the old one before running tests again. This is something that should be automated at some point.


browser-tests/registrationView.ts
```diff
@@ -0,0 +1,95 @@
...
+test('Test all required fields if user is minor', async t => {
+  await login(t, 'minor');
+
+  await t.click(registrationFormSelector.submitButton);
+
+  const allRequiredErrors = await Selector(
+    'div[class^="TextInput-module_helperText"]'
+  ).count;
+
+  await t.expect(allRequiredErrors).eql(10);
```
Jokinen, line 31: Minor comment. Could even a similar check to `errorCount > 0` be enough here? This looks good also. I'm thinking about a scenario where we add or remove fields--those would break this test without necessarily breaking the functionality.

Jokinen (approved):
Looks good to me, good job 👍

I left one minor comment.

browser-tests/registrationView.ts
tuovinensanttu: I guess it could be enough. My main question is do we only wan't to test that user can't submit the form straight up, or do we want to test that correct amount of fields are marked as required, based on user's age.

Jokinen: Hmn, really good point 🤔 In that case I would maybe leave it as is.

One approach would be to fill the fields that are normally required, and then observe that the submission still fails. But I'm not really sure if that's better. I'd leave it and make changes later is any problems arise.

PR #162, opened 2020-09-23
YM-362 | Fix scroll to first error

tuovinensanttu:
Scroll to first error was broken. When getting error keys, the order would be pretty much random. I added logic that sorts keys to match the order in which the fields occur in our form. 

Also added scroll handling to `additionalContactPersons`. After #160 is merged, I'll try to write e2e test so the functionality won't break again.

### DIFF
--- a/src/domain/youthProfile/form/FormikFocusError.tsx
+++ b/src/domain/youthProfile/form/FormikFocusError.tsx
@@ -19,6 +19,19 @@ const defaultConfig: ScrollConfig = {
   duration: 800,
 };
 
+const FIELD_ORDER = [
+  'firstName',
+  'lastName',
+  'primaryAddress',
+  'phone',
+  'approverFirstName',
+  'approverLastName',
+  'approverEmail',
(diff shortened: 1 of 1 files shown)


Jokinen (approved):
This looks good to me and I'd merge it as it.

Something I'm wondering is if we could use the DOM order to find the first error. The benefit would be that we don't have to maintain an array of fields that duplicates the DOM order. Something that supports that approach is the default form behaviour for instance in Chrome that focuses the first incorrect field. In practice it could happen for instance by looping over the [form elements](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement/elements), but I don't know what difficulties could arise.

tuovinensanttu:
> Something I'm wondering is if we could use the DOM order to find the first error in the next iteration. 

I gave this a little test, and I think it would be possible. Maybe this is something that could be done in the future, at the moment this solution seems to be pretty prone to errors and `types` toss a nice twist into the mix as well.

PR #163, opened 2020-09-23
YM-362 | Add e2e test for scrolling to first error

tuovinensanttu:
Added `2e2` test for testing "scroll to first error element" functionality. It's added to both "check errors" tests, it will take extra second to execute so I think it should be fine.

I just realised that I haven't remembered to update `CHANGELOG.md`. I included those to this PR as well.


Jokinen (approved):
Looks good 👍

PR #164, opened 2020-09-23
YM-354 | Prevent double submitting form

tuovinensanttu:
After this change, even if user spams "Register" button, `createProfile` - mutation is only executed once. `loading` wasn't enough alone, so I added a state that is changed after profile is created.


Jokinen (approved):
As far as I can understand, this approach should work and it looks good to me.

My first instinct would be to put this logic into the form component itself. That way the editing case would also be covered (even though it may not cause an error, it's still unnecessary to send the same request multiple times). Did you try that?

tuovinensanttu:
> My first instinct would be to put this logic into the form component itself. 
🤦  for some reason that did not occur to me. I moved logic inside formik's `onSubmit` handler.

PR #166, opened 2020-09-28
YM-358 | Update meta description

Jokinen:
Updates the index.html file to contain the updated meta description.

I also slightly refactored the `PageWrapper`. Application language is now synced closer to root and `PageWrapper` only contains meta tags that could, in the future, be page specific.


tuovinensanttu (approved):
Looks good 👍

PR #169, opened 2020-10-02
YM-379 | View own information tests

tuovinensanttu:
## Description
- Added `e2e` - tests that navigate to "Profile information" page and checks some of the entered values (if needs be this can be extended to cover all information user entered. Relating to back-end separation, would this actually be a good idea?).

## Context
Tests are always good 😄 

User birth year values are no longer hardcoded. This will prevent tests from crashing in the distant future.

I also created `.testcaferc.json` file. This was necessary to alter the order in which testcafe runs the tests, as one would expect, it uses alphabetical order as default. Moving file order to different file should also keep `package.json` script section little cleaner. I decided to leave browser information inside `browser-test` scripts for now.

[YM-379](https://helsinkisolutionoffice.atlassian.net/browse/YM-379)

## How Has This Been Tested?
`yarn browser-test` -> all test pass.

1) Add values to `.env` file (if you don't have these you can contact me):
- `REACT_APP_TESTING_USERNAME`
- `REACT_APP_TESTING_PASSWORD`
2) Login to [profile-admin](https://profiili-api.test.kuva.hel.ninja/admin) and make sure "Uno User" has no profile
3) Start local application `yarn start`
4) Run `yarn browser-test`
5) Delete the created profile to avoid crashing next time tests are run.


.testcaferc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "src": [
+    "browser-tests/loginView.ts", "browser-tests/registrationView.ts", "browser-tests/membershipView.ts"
```
Jokinen, line 3: I have two big question marks with this approach

1. `src` setting makes no mention of being a tool for controlling execution order: https://devexpress.github.io/testcafe/documentation/reference/configuration-file.html#src
2. With TestCafe, and tests in general, it's recommended that test cases don't depend on each other.

That's why I'm hesitant to accept this change. We discussed this in more detail on Slack and didn't find a clear winning formula. The best possible alternative was one Santtu came up with and it's to use multiple test accounts that are in different states of their "lifecycle". This removes the dependency, but increases the maintenance load. I'd cautiously vote for this approach, because it allows us to avoid doing 1 or 2.

Jokinen (approved):

.testcaferc.json
tuovinensanttu: I'll go with the multiple profile approach, after sleeping on this it makes much more sense.

Jokinen (approved):
This looks good now in my opinion. Thank you for taking the time to adjust the implementation 👍

PR #170, opened 2020-10-05
YM-376 | GraphQL error: You don't have permission

tuovinensanttu:
## Description

Changed `HasYouthProfile` query to use lazy query instead. Query is executed after we make sure user is authenticated.

## Context

When user entered website, it would immediately trigger `HasYouthProfile` query and cause `[GraphQL error]`. Now that query is only executed after user is authenticated, error shouldn't exist anymore.

[YM-376](https://helsinkisolutionoffice.atlassian.net/browse/YM-376)

## How Has This Been Tested?

Changes has only tested manually by me.

1) With `development` branch, start local environment with `yarn start`.
2) When process is running, enter the site again via address bar. You should see `[GraphQL error]: You don't have permissions` error in the browser console (refreshing won't trigger the error).
3) Change branches and repeat steps 1 & 2. You should not see the previously mentioned error.

### DIFF
--- a/src/domain/app/AppYouthProfileRoute.tsx
+++ b/src/domain/app/AppYouthProfileRoute.tsx
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
 import { Route, Redirect, RouteProps } from 'react-router';
 import { useLazyQuery } from '@apollo/react-hooks';
 import { loader } from 'graphql.macro';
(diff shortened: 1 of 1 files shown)


src/domain/app/AppYouthProfileRoute.tsx
```diff
@@ -19,13 +21,25 @@ type Props = RouteProps;
...
+      onError: (error: Error) => {
+        Sentry.captureException(error);
+        setShowNotification(true);
+      },
+    }
+  );
+
+  const isAuthenticated = useSelector(isAuthenticatedSelector);
+
+  if (isAuthenticated && !profileLoaded) {
```
Jokinen, line 39: Should this be inside a `useEffect`?

Jokinen (approved):
I tested the changes and could not longer see an error.

src/domain/app/AppYouthProfileRoute.tsx
tuovinensanttu: Changed to `useEffect`

PR #173, opened 2020-10-07
YM-385 | Routing is broken

tuovinensanttu:
## Description

Reverted previous changes. `useQuery` hook is back and errors related to permission are ignored.

## Context

Changes are necessary to restore correct routing. 

[YM-385](https://helsinkisolutionoffice.atlassian.net/browse/YM-385)

## How Has This Been Tested?

Tested profile creation manually and with `e2e` - tests. Tests do now cover 100% functionality yet, but they are better than nothing.

If application is opened in incognito for example, permission error is show but it shouldn't crash the application.

### DIFF
--- a/src/domain/app/AppYouthProfileRoute.tsx
+++ b/src/domain/app/AppYouthProfileRoute.tsx
@@ -24,9 +24,10 @@ function AppYouthProfileRoute(props: Props) {
 
   const { data, loading } = useQuery<HasYouthProfile>(HAS_YOUTH_PROFILE, {
     onError: error => {
-      // Ignore permission errors. This decision was made based on the difficulty to add enough checks that slow down Redirecting.
-      // Using useLazyQuery with isAuthenticatedSelector won't fetch profile information fast enough and user is always
-      // redirected to LoginPage. If in the future permission error causes application to crash, some kind of custom hook is necessary.
+      // It's too difficult to block the profile from being loaded before authentication is finished.
+      // Instead we allow the UI to optimistically request the profile before it has the authorisation to do so.
+      // Because we ignore the subsequent permissions error, the user experience is not affected.
+      // Other errors should be shown to the user and be passed to Sentry.
       error.graphQLErrors.forEach(qlError => {
         if (qlError.message !== authConstants.PERMISSION_DENIED) {
           Sentry.captureException(qlError);
--- a/src/domain/app/AppYouthProfileRoute.tsx
+++ b/src/domain/app/AppYouthProfileRoute.tsx
@@ -1,17 +1,16 @@
-import React, { useState, useEffect } from 'react';
+import React, { useState } from 'react';
 import { Route, Redirect, RouteProps } from 'react-router';
-import { useLazyQuery } from '@apollo/react-hooks';
+import { useQuery } from '@apollo/react-hooks';
 import { loader } from 'graphql.macro';
 import * as Sentry from '@sentry/browser';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


src/domain/app/AppYouthProfileRoute.tsx
```diff
@@ -23,21 +22,19 @@ function AppYouthProfileRoute(props: Props) {
...
-  );
-
-  const isAuthenticated = useSelector(isAuthenticatedSelector);
-
-  useEffect(() => {
-    if (isAuthenticated) loadProfile();
-  }, [isAuthenticated, loadProfile]);
+  const { data, loading } = useQuery<HasYouthProfile>(HAS_YOUTH_PROFILE, {
+    onError: error => {
+      // Ignore permission errors. This decision was made based on the difficulty to add enough checks that slow down Redirecting.
```
Jokinen, line 27: I know what you mean here because we had a chat on Slack, but my guess is that people without that context might find it difficult to understand what slowing down redirecting means. It's not an easy concept to write in simple terms so I have attempted to give another perspective:

> It's too difficult to block the profile from being loaded before authentication is finished. Instead we allow the UI to optimistically request the profile before it has the authorisation to do so. Because we ignore the subsequent permissions error, the user experience is not affected. Other errors should be shown to the user and be passed to Sentry.

src/domain/app/AppYouthProfileRoute.tsx
```diff
@@ -23,21 +22,19 @@ function AppYouthProfileRoute(props: Props) {
...
+      // Using useLazyQuery with isAuthenticatedSelector won't fetch profile information fast enough and user is always
+      // redirected to LoginPage. If in the future permission error causes application to crash, some kind of custom hook is necessary.
+      error.graphQLErrors.forEach(qlError => {
+        if (qlError.message !== authConstants.PERMISSION_DENIED) {
+          Sentry.captureException(qlError);
```
Jokinen, line 32: You are passing the error to Sentry within a loop. A lot of the time we only pass the error object once

```tsx
onError: (error: Error) => {
  Sentry.captureException(error);
  // ...rest
}
```

I don't actually know which approach is more correct, because I don't have an idea of how Sentry behaves in the two cases. I just want to point this out.

Jokinen (approved):
Minor comments.

src/domain/app/AppYouthProfileRoute.tsx
tuovinensanttu: If I pass down it like that, permission errors are sent as well and thats what I'm trying to prevent. I find it kind of impossible that there would be more than one error (now that permission is excluded that is).

src/domain/app/AppYouthProfileRoute.tsx
tuovinensanttu: That makes much more sense, thanks for writing this. I have a bad habit to write first thing that comes to mind and in most my massage is hard to understand 😅

src/domain/app/AppYouthProfileRoute.tsx
Jokinen: This is a pattern I used in another project:

```tsx
onError: (error: Error) => {
  if (!includesError(error.graphQLErrors, permissionError)) {
    Sentry.captureException(error);
    // ...rest
  }
}
```

But I don't think that you should spend time on changing this. If I needed to guess, I'd say that your approach is the more correct one. Maybe I wrote this down, because it's more of a mental note for me that this is something I should keep in mind when implementing error handling in the future.

PR #178, opened 2020-10-16
YM-405 | Centralize Apollo and Sentry integration

Jokinen:
## Description

Previously the Apollo and Sentry integration was done multiple times at
multiple different places in the code. This commit centralizes this
logic by logging errors into Sentry by using the 'onError' field in
Apollo boost's client.

This centralized error handling adds more context to the data we pass
on to Sentry. This makes debugging easier by allowing us to see the
name of the query that caused an error, for instance. This kind of
context couldn't have been added without centralization.

This change also implements a centralized error ignoring mechanism.
We know that we do not want to pass some errors to Sentry, because the
errors we receive are expected. Such error often include for instance
permission related errors. Now such errors can be ignored globally,
making it less likely that we forget to ignore them in some view.

Errors can be ignore or consumed. Ignored errors won't be sent to
Sentry, but they will be sent to consumers like 'useQuery' hooks. When
an error is consumed, it won't be passed on to consumers, in which case
'useQuery' hook's 'onError' callback would not be called.

Note that when consuming en error, the entire error object is consumed.
This means that if one error in an error object containing two errors
is consumed, both errors are consumed.

Note that global error consuming should be used scarcely. It should be
reserved for those errors of whose exclusion the view layer clearly
benefits from.

The following changes were made into the data that is sent into Sentry
- GraphQL errors are now sen as errors instead of info
- The following tags were added (can be used to search)
    - GraphQL Errors
        - *type:* Always with the value `GraphQL Error`
        - *operation.name:* The name of the query or mutation that caused the error
        - *operation.kind:* The kind of operation that caused the error (`query`/`mutation`/ `subscription`)
   - Network Error
        - *type:* Always with the value `Network Error`
- The following contexts were added into GraphQL errors
    - GraphQL Error
         - *extensions:* Stringified JSON of extensions. May contain, for instance, the error code.
         - *locations:* Stringified JSON of the location of the error.
         - *path:* Stringified JSON of the path of the error.
    - Operation
         - *name:* Name of operation.
         - *kind:* The kind of the operation.

## Context
I set out to fix YM-405. Because the errors in Sentry were a bit vague, I looked into how to add more contextual data. I found Apollo's documentation for error handling, which guides towards using the `onError` prop: https://www.apollographql.com/docs/react/data/error-handling/. I felt that this extra context was useful enough to have to warrant some refactoring.

This change also fixes another pet-peeve I've had with this application--Sentry logging is done on bit all over the place. With these changes, Sentry integrates into Apollo in one place. I removed all the Sentry calls I knew were redundant. However, some duplicate logging may have remained in mutation calls. Some of these  had the potential to catch errors that were not Apollo related so I left them in. 

[YM-405](https://helsinkisolutionoffice.atlassian.net/browse/YM-405)

## How Has This Been Tested?
I've added unit tests for the `handleError` function. These rather ruthlessly mock `Sentry` and `configs`. The fix I saw for this was to parameterise the code I was mocking, but I felt that it was a good tradeoff to avoid doing that for now.

I also logged events from my local installation into Sentry with a special environment name and release name so that I could easily clean them up. I reproduced the error YM-405 was about and debugged it by adding more context. I then checked that those errors were no longer logged into Sentry after I had ignored them.

As a non logged in user

1. Go to `/create`
2. Go to Sentry and verify that no error was generated there

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,10 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
+## [Unreleased]
+### Fixed
+- Logging uninteresting error messages to Sentry
+
 ## [1.1.1] - 2020-10-07
 ### Added
 - e2e - tests for viewing own information
--- a/src/domain/app/AppYouthProfileRoute.tsx
+++ b/src/domain/app/AppYouthProfileRoute.tsx
@@ -2,7 +2,6 @@ import React, { useState } from 'react';
 import { Route, Redirect, RouteProps } from 'react-router';
 import { useQuery } from '@apollo/react-hooks';
 import { loader } from 'graphql.macro';
-import * as Sentry from '@sentry/browser';
 import { useTranslation } from 'react-i18next';
 import { isValid, parseISO } from 'date-fns';
 
@@ -10,7 +9,6 @@ import { HasYouthProfile } from '../../graphql/generatedTypes';
 import getCookie from '../../common/helpers/getCookie';
 import LoadingContent from '../../common/components/loading/LoadingContent';
 import NotificationComponent from '../../common/components/notification/NotificationComponent';
-import authConstants from '../auth/constants/authConstants';
 
 const HAS_YOUTH_PROFILE = loader(
   '../youthProfile/graphql/HasYouthProfile.graphql'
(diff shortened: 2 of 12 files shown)


tuovinensanttu (approved):
This looks great! I was actually thinking doing something similar with [YM-376](https://helsinkisolutionoffice.atlassian.net/browse/YM-376), but seems that you beat me to it 😄

PR #183, opened 2020-11-03
YM-389 | Change registration to work with federation

tuovinensanttu:
## Description
Profile creation now works with federation. 
Fixed eslint warnings that I missed last time.

## Context

`-`

[YM-389](https://helsinkisolutionoffice.atlassian.net/browse/YM-389)

## How Has This Been Tested?
This has only been tested manually for now.

1) Navigate localhost
2) Enter birthdate and login
3) Fill the registration and submit form
4) Expect to land either on "Application send" or "Membership information" page

### DIFF
--- a/src/domain/youthProfile/create/CreateYouthProfile.tsx
+++ b/src/domain/youthProfile/create/CreateYouthProfile.tsx
@@ -32,7 +32,9 @@ function CreateYouthProfile({
 }: Props) {
   const [showNotification, setShowNotification] = useState<boolean>(false);
   const { i18n } = useTranslation();
-  const [createProfiles, { loading }] = useCreateProfiles(setShowNotification);
+  const [createProfiles, { loading }] = useCreateProfiles({
+    onError: () => setShowNotification(true),
+  });
 
   const birthDate = getCookie('birthDate');
   const isBirthDateValid = isValid(parseISO(birthDate));
--- a/src/domain/youthProfile/create/CreateYouthProfile.tsx
+++ b/src/domain/youthProfile/create/CreateYouthProfile.tsx
@@ -1,48 +1,25 @@
 /* eslint-disable sort-keys */
-import React, { useState } from 'react';
+import React from 'react';
 import { User } from 'oidc-client';
-import { useMutation } from '@apollo/react-hooks';
-import { loader } from 'graphql.macro';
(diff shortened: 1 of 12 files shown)


src/domain/youthProfile/create/CreateYouthProfile.tsx
```diff
@@ -51,12 +57,18 @@ function CreateYouthProfile({
...
   const history = useHistory();
+  const profileApiToken = useSelector(profileApiTokenSelector);
 
-  const [createProfile, { loading }] = useMutation<
+  const [createProfile, { loading: creatingProfile }] = useMutation<
     CreateMyProfileData,
     CreateMyProfileVariables
   >(CREATE_PROFILE);
 
+  const [createMyYouthProfile, { loading: creatingYouthProfile }] = useMutation<
```
Jokinen, line 67: This approach works and is fine, but I'd consider whether it'd make sense to encapsulate it more. You could for instance have a custom hook which hides the two mutations calls behind a single call. This is a flaky spot in the architecture of this application in some sense, because we are integrating two APIs and encapsulation would provide some level of protection against regressions.

Jokinen (approved):
I tested the UI according to the instructions and was able to reproduce the results 👍

src/domain/youthProfile/create/CreateYouthProfile.tsx
tuovinensanttu: Good idea, I'll create a custom hook for this 👍

src/domain/youthProfile/hooks/useCreateProfiles.ts
```diff
@@ -0,0 +1,154 @@
...
+          if (result.data) createYouthProfile(formValues);
+        })
+        .catch((error: Error) => {
+          Sentry.captureException(error);
+          setError(error);
+        });
+    }
+  };
+
+  return { createProfiles, isLoading, error, setError };
```
Jokinen, line 151: I think you should use the same signatures as Apollo hooks have.

`[mutationFunction, { loading, error }]`

I would not make `setError` public. I think the error this hook has should represent the possible errors in the mutations.

I would consider whether the returned error should not be derived from the errors in the two mutation hooks directly without storing them within another state. This is a more minor point imo.

PR #184, opened 2020-11-06
YM-391 | Fix approve page profile fetching with federation

Jokinen:
## Description

Implements a `useProfileByTokens` hook that can be used to fetch a combination of helsinki profile and youth profile matching the profile that's returned from for instance `myYouthProfile` query.

I addressed new possible error codes on the `ApproveYouthProfilePage` and refactored it so that's it's easier to understand.

I also refactored the `getAddress` and `getAddresses` helpers which were entangled with the old types.

## Context

After this change guardians are able to approve youths against the federated API.

[YM-391](https://helsinkisolutionoffice.atlassian.net/browse/YM-391)

## How Has This Been Tested?

I added unit tests to the `useProfileByTokens` hook even though it's a bit questionable whether it's useful to test it independently as opposed as a part of some view.

I also added tests into the the `ApproveYouthProfilePage` component which check that it returns the expected error view when it faces some of the new error codes.

I ran the browser tests locally and they looked to be working as expected. However, because some mutations are not yet handled in this PR, the tests couldn't be completed, which means that there's not certain confidence that the approval view would pass them all.

1. Create a new youth profile (makes sure you create it to the federated application)
1. Open the approve link in a youth-membership-ui based on this branch
1. Expect to see all profile data
1. Approve profile
1. Expect to be redirected into the success view without errors


tuovinensanttu (approved):
Looks good to me. Approving profile was a success! Nice work!

PR #186, opened 2020-11-10
YM-390 | Edit mutations work with federation

tuovinensanttu:
## Description

This PR changes resend email, renew membership and edit profile information mutations to work with federation. 

## Context

`-`

[YM-390](https://helsinkisolutionoffice.atlassian.net/browse/YM-390)

## How Has This Been Tested?

Manually + `e2e`-tests (note that this branch is not up to date with development, where profile creation tests are skipped)

1) Create a youth profile (adult, emails are not yet received)
2) Editing youth profile data works
3) From jässäris django admin set expiration year to 2020
4) Expect renew button to show and pressing it renews membership succesfully

### DIFF
--- a/src/common/test/testUtils.tsx
+++ b/src/common/test/testUtils.tsx
@@ -3,6 +3,7 @@ import { Provider } from 'react-redux';
 import { mount, shallow } from 'enzyme';
 import { act } from 'react-dom/test-utils';
 import { MockedProvider, MockedResponse } from '@apollo/react-testing';
+import { MemoryRouter } from 'react-router';
 
 import store from '../../redux/store';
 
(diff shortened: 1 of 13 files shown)


src/common/test/testUtils.tsx
```diff
@@ -21,6 +21,18 @@ export const mountWithApolloProvider = (
       {children}
     </MockedProvider>
   );
+
+export const mountWithApolloAndReduxProviders = (
```
Jokinen, line 25: By default, I would have added the `<Provider>` into the current extended `mount` operation. In the long term it wouldn't make sense to have a new flavour whenever a new provider needs to be added.

If you wish, you can also consider the pattern shown here:

https://enzymejs.github.io/enzyme/docs/api/ReactWrapper/getWrappingComponent.html#examples

It would allow you to separate and share the `Provider` logic.

src/domain/youthProfile/edit/EditYouthProfile.tsx
```diff
@@ -43,28 +38,14 @@ function EditYouthProfile(props: Props) {
...
-
-    updateProfile({ variables })
-      .then(() => {
-        history.push('/membership-details');
-      })
-      .catch((error: Error) => {
-        Sentry.captureException(error);
-        setShowNotification(true);
-      });
+    updateProfiles(formValues, data);
```
Jokinen, line 48: I think having the redirect in this level is better separation of concerns. Some consumers of the hook may want to redirect somewhere else than this view does. Generally, doing the redirect here allows the hook to be agnostic about the url structure of this application for instance.

src/domain/youthProfile/hooks/useUpdateProfiles.ts
```diff
@@ -0,0 +1,80 @@
...
+      input: {
+        youthProfile: getEditYouthProfile(formValues, membershipData),
+        profileApiToken,
+      },
+    };
+
+    try {
+      await updateMyProfile({ variables: myProfileVariables });
+      await updateMyYouthProfile({ variables: myYouthProfileVariables });
+      history.push('/membership-details');
```
Jokinen, line 67: I would move this to an `onCompleted` prop or I would wait for `updateProfiles` in the consuming code and then make the redirect.

Jokinen (approved):
Looked good 👍 A few minor comments.

I wasn't able to get a confirmation message to pass so I wasn't able to test these.

src/domain/youthProfile/hooks/useUpdateProfiles.ts
tuovinensanttu: Good point, I'll move this up a step.

src/common/test/testUtils.tsx
tuovinensanttu: I decided just to compine all provers under single helper. I tried to logic described in the docs, but it resulted to apollo network error.

PR #197, opened 2021-01-13
YM-444 | Fix failing browser test

Jokinen:
## Description

The test data created a user that will be over 29 years after march, but assumed them to be over 30 already.

As a dirty fix I just changed their birthday into 01.01. Another approach would have been to find out the date that an exactly 30-year-old would be, but I am not sure if that would have been much better, and it would have required more work to implement.

## Context

[YM-444](https://helsinkisolutionoffice.atlassian.net/browse/YM-444)

## How Has This Been Tested?

I ran browser test against this branch: https://github.com/City-of-Helsinki/youth-membership-ui/actions/runs/482027598

Check that browser test have passed here: https://github.com/City-of-Helsinki/youth-membership-ui/actions/runs/482027598

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Fixed
 
 - Year dependent failing snapshot test
+- Browser test that would always fail until march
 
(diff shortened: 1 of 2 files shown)


tuovinensanttu (approved):
Test works now 👍

PR #198, opened 2021-01-13
Chore | Fix failing snapshot test

Jokinen:
## Description

After the year changed, the test data no longer represent an user of
the expected age. This is a quick fix that changes the user's birthday
so that they are the correct age.

The problem with this fix is that the snapshot test will break once the year changes again. The test should ideally be refactored so that it's not dependent on current time and would not break as time moves onwards.

## How Has This Been Tested?

`yarn test` passes without errors.

Ensure that CI tests pass.


tuovinensanttu (approved):
All tests pass 👍

PR #199, opened 2021-01-14
Chore | Upgrade HDS 0.11.3 → 0.20.0

Jokinen:
## Description

Upgrades HDS to the latest version.

The latest version includes for instance the navigation component we can adopt to fix accessibility issues in our mobile navigation.

Since 0.11.3 there have been some breaking changes: https://github.com/City-of-Helsinki/helsinki-design-system/blob/master/CHANGELOG.md

Some design tokens were renamed and the notification component was re-written. I updated the implementation for the notification component.

Here are the design tokens I updated:
- `fontsize-h-1` to `fontsize-heading-xl`
- `fontsize-h-2` to `fontsize-heading-l`
- `fontsize-h-3` to `fontsize-heading-m`
- `fontsize-h-4` to `fontsize-heading-s`
- `fontsize-body-small` to `fontsize-body-s`
- `fontsize-body-default` to `fontsize-body-m`
- `fontsize-body-medium` to `fontsize-body-l`
- `fontsize-body-large` to `fontsize-body-xl`
- `--color-coat-of-arms-blue` to `coat-of-arms`

Global styles were removed from HDS. Our non-HDS buttons relied on the reset styles that HDS applied for styling. I added reset styles for buttons into the repo. I also changed `box-sizing` to be `border-box` by default--something HDS's global styles did before.

Due to style load order some heading styles were being overwritten in a way which diverged from the design after the upgrade. It's why I also had to apply most headings by using the `<Text />` component.

Finally I had to adjust the implementation of the `<TextInput />` component in order to make it work without visual glitches in a grid layout. I also refactored the implementation to take the changed API into account.

## How Has This Been Tested?

I've tested manually by browsing through some of the views. I've checked that the test suite passes.

I can't create concise instructions for testing this ticket as HDS is everywhere in the app. I personally had the test version and a version against these changes open in the same window and I then jumped between them in order to spot changes in the layout.


package.json
```diff
@@ -15,7 +15,7 @@
     "@types/jest": "24.0.18",
     "@types/lodash": "^4.14.151",
     "@types/node": "^13.9.8",
-    "@types/react": "16.9.3",
+    "@types/react": "16.9.12",
```
Jokinen, line 18: I had to update this type library, because the older one didn't seem to include the general attribute `translate`, which resulted in that attribute being required for many HDS components. I assume that HDS's npm package was built with a newer version of this type library that does include the `translate` attribute.

src/common/components/select/Select.module.css
```diff
@@ -1,56 +1,61 @@
 .select {
-    display: flex;
-    flex-direction: column;
-    width: 100%;
+  display: flex;
```
Jokinen, line 2: Apologies for the whitespace changes here. I used my IDE's replace tool so I did not notice that whitespace had been changed.

src/domain/auth/components/login/Login.tsx
```diff
@@ -61,6 +62,7 @@ function Login(props: Props) {
                     onClick={authenticate}
                     variant="supplementary"
                     className={styles.button}
+                    iconRight={null}
```
Jokinen, line 65: Newer versions of HDS require `iconLeft`/`iconRight` for supplementary buttons. For now I avoided design changes by using `null`.

tuovinensanttu (approved):
Found nothing to comment on, solid work!

PR #200, opened 2021-01-14
FIX | Forward ref in TextInput integration

Jokinen:
## Description

The wrapper I added in #199 did not forward ref, which caused an error to be logged in the birthday input component.

1. Open the application without being logged in
2. Expect no ref error to be logged


tuovinensanttu (approved):
Error is gone from the console 👍 . While testing this I noticed that birthdate inputs error message is wrong color.

image

Jokinen:
Good catch, I'll make a PR for that as well.

PR #202, opened 2021-01-14
FIX | Fix incorrect error colour in DateInput

Jokinen:
## Description

The hds-core package had changed the way in which it applies error states. This PR changes the DateInput component to use that approach.

## How Has This Been Tested?

I've tested this manually.

1. Input a date that causes an error
2. Expect to see an error message

### DIFF
--- a/src/common/components/dateInput/DateInputGroup.tsx
+++ b/src/common/components/dateInput/DateInputGroup.tsx
@@ -15,11 +15,13 @@ interface Props {
 
 function DateGroup({ children, id = INPUT_GROUP_ID, label, error }: Props) {
   const isInvalid = Boolean(error);
+  const errorId = `${id}-error`;
 
   return (
     <div
       role="group"
       aria-labelledby={id}
+      aria-describedby={errorId}
       className={className({
         'hds-text-input': true,
         'hds-text-input--invalid': isInvalid,
@@ -27,7 +29,11 @@ function DateGroup({ children, id = INPUT_GROUP_ID, label, error }: Props) {
     >
       <DateInputLabel id={id}>{label}</DateInputLabel>
       {children}
-      <span className="hds-text-input__helper-text">{error}</span>
+      {isInvalid && (
+        <span id={errorId} className="hds-text-input__error-text">
+          {error}
+        </span>
+      )}
     </div>
   );
 }
--- a/src/common/components/dateInput/DateInputGroup.tsx
+++ b/src/common/components/dateInput/DateInputGroup.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-import className from 'classnames';
 import 'hds-core/lib/components/text-input/text-input.css';
 
 import DateInputLabel from './DateInputLabel';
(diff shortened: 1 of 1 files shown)


src/common/components/dateInput/DateInputGroup.tsx
```diff
@@ -15,19 +15,25 @@ interface Props {
...
+  const errorId = `${id}-error`;
 
   return (
     <div
       role="group"
       aria-labelledby={id}
+      aria-describedby={errorId}
       className={className({
         'hds-text-input': true,
         'hds-text-input--invalid': isInvalid,
```
Jokinen, line 27: Maybe this could be removed as well.

src/common/components/dateInput/DateInputGroup.tsx
```diff
@@ -15,19 +15,25 @@ interface Props {
 
 function DateGroup({ children, id = INPUT_GROUP_ID, label, error }: Props) {
   const isInvalid = Boolean(error);
+  const errorId = `${id}-error`;
 
   return (
     <div
       role="group"
       aria-labelledby={id}
+      aria-describedby={errorId}
```
Jokinen, line 24: This doesn't do much, but I sneaked it in here. The error will be read when the user focuses the group.

src/common/components/dateInput/DateInputGroup.tsx
```diff
@@ -15,19 +15,25 @@ interface Props {
...
       className={className({
         'hds-text-input': true,
         'hds-text-input--invalid': isInvalid,
       })}
     >
       <DateInputLabel id={id}>{label}</DateInputLabel>
       {children}
-      <span className="hds-text-input__helper-text">{error}</span>
+      {isInvalid && (
```
Jokinen, line 32: The new error styles contain an icon so the error element has to be hidden until there's an error.

src/common/components/dateInput/DateInputGroup.tsx
tuovinensanttu: I didn't notice this would affect anything, tested with voice over + safari / chrome

src/common/components/dateInput/DateInputGroup.tsx
tuovinensanttu: Seems to work without it.

tuovinensanttu (approved):
LGTM!

src/common/components/dateInput/DateInputGroup.tsx
Jokinen: You will have to use the <kbd>ctrl</kbd> + <kbd>opt</kbd> + <kbd>right arrow</kbd> combination so that you'll move to the group. You also have to stay on the group element for a moment for the description to be read. Still, I doubt it helps much in practice.

tuovinensanttu: Ahh of course! For some reason when I'm using keyboard I just tab myself through everything 😄

src/common/components/dateInput/DateInputGroup.tsx
Jokinen: I removed this class.

PR #203, opened 2021-01-15
YM-404 | Fix duplicate cancel button while editing own information

tuovinensanttu:
## Description

Replaced the `<Link> + <button>` combo with `LinkButton`, this removed excess button element.

## Context

Duplicate elements would cause confusion to users with screen reader.

[YM-404](https://helsinkisolutionoffice.atlassian.net/browse/YM-404)

## How Has This Been Tested?

Tested this manually with voice over + chrome.

1. While using screen reader, navigate to "Edit own information page"
2. Move to cancel button
3. Expect button text to be read once

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Year dependent failing snapshot test
 - Browser test that would always fail until march
+- While editing own information, there was "duplicate" cancel button. 
 
 ## [1.2.1] - 2020-11-25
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 - Year dependent failing snapshot test
 - Browser test that would always fail until march
+- While editing own information, there was "duplicate" cancel button.
```
Jokinen, line 18: How about

>- Duplicate cancel button in information editing view

tuovinensanttu: True, that makes more sense. I was staring this line like 5mins and that was the best I came up with 😅

Jokinen (approved):
Tested the change and it worked!

PR #204, opened 2021-01-15
YM-398 | Pages don't have unique titles

tuovinensanttu:
## Description

Added unique titles to all pages. At the start, I made a wrong move and tried to add title changer to `PageContent`. After I realised that titles were in the `PageWrapper` I moved everything the there. Not really sure that should I revert the `title` props that are passed to `PageContent`, although they should help with analytics...

## Context

Previously page title was always "Jässäri", this caused confusion for screen reader users, because they couldn't tell if new page was loaded succesfully or not.

[YM-398](https://helsinkisolutionoffice.atlassian.net/browse/YM-398)

## How Has This Been Tested?
I have tested this manually

1. Open the web page.
2. Expect title to change on every navigation (expect login page if user age is less than 13).
3. Expect screen reader to read correct title.


Jokinen (changes requested):
Linting is broken.

src/domain/app/AppNonProfileRoute.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import AppPageTitleRoute from './AppPageTitleRoute';
+import { RouteProps } from 'react-router';
+
+type Props = RouteProps & {
+  pageTitle: string;
+};
+
+const AppNonProfileRoute = (props: Props) => {
```
Jokinen, line 10: Do you need to have this compnent or could you just directly use `AppPageTitleRoute`?

src/domain/app/AppNonProfileRoute.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+
+import AppPageTitleRoute from './AppPageTitleRoute';
+import { RouteProps } from 'react-router';
```
Jokinen, line 4: Imports from `node_modules` should come before in-app imports based on the linting rules in this project as far as I know. Maybe your IDE didn't integrate with eslint successfully?

src/domain/app/AppPageTitleRoute.tsx
```diff
@@ -0,0 +1,24 @@
...
+const AppPageTitleRoute = (props: Props) => {
+  const { t } = useTranslation();
+
+  return (
+    <>
+      <Helmet>
+        <title>{t(props.pageTitle)}</title>
+        <meta property="og:title" content={t(props.pageTitle)} />
+      </Helmet>
+      <Route {...props} />
```
Jokinen, line 19: Did you notice that you are passing `pageTitle` to `<Route />`?

I recommend that you use the destructuring pattern:

`({ pageTitle, ...rest })` and pass `...rest` to `<Route />`.

Jokinen (approved):

src/domain/app/AppNonProfileRoute.tsx
tuovinensanttu: It works without, I'll remove it. For some reason my brain thought that I need another component for these 😄

src/domain/app/AppPageTitleRoute.tsx
tuovinensanttu: Oops, I'll fix this.

Jokinen:
Nice work overall 👍 This was a more difficult fix than I anticipated. Well done.

PR #205, opened 2021-01-18
YM-399 | Fix links that don't warn when opening a new window

Jokinen:
## Description

Replaces the links that open a new window with the "ExternalLink" component which includes an icon warning for visual users and a text warning for screen reader users.

## Context

[YM-399](https://helsinkisolutionoffice.atlassian.net/browse/YM-399)

## How Has This Been Tested?

I've tested the feedback link and the links within the terms field manually with Safari and VoiceOver and I've added unit tests for the `ExternalLink` component.

With a screen reader

1. Navigate to feedback link (footer area)
2. Expect to hear a warning about the link opening in a new tab

.

1. Navigate to user creation form
2. Expect all the links within the terms field label to warn that they open in a new tab

Note that labels containing links are not good for accessibility, but it's beyond the scope of this change.

### DIFF
--- a/src/common/components/linkButton/LinkButton.tsx
+++ b/src/common/components/linkButton/LinkButton.tsx
@@ -25,7 +25,7 @@ function LinkButton({
   rel,
   iconRight,
 }: Props) {
-  const iconRightWithPadding = iconRight ? ` ${iconRight}` : '';
+  const iconRightWithPadding = iconRight ? <> {iconRight}</> : '';
 
   if (component === 'Link') {
     return (
--- a/src/common/components/linkButton/LinkButton.tsx
+++ b/src/common/components/linkButton/LinkButton.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 import classNames from 'classnames';
 import { Link } from 'react-router-dom';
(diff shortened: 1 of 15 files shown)


tuovinensanttu:
There should be few links / buttons in "under age of 13 page" (e.g. birthdate is 1.1.2010). One opens a pdf in new window and the other one opens servicemap.

Jokinen:
Good point. I've been in contact with the designer to verify the design.

src/common/components/linkButton/LinkButton.tsx
```diff
@@ -22,14 +23,18 @@ function LinkButton({
   component,
   variant,
   rel,
+  iconRight,
 }: Props) {
+  const iconRightWithPadding = iconRight ? ` ${iconRight}` : '';
+
```
tuovinensanttu, line 29: If the `iconRight` is made a string, it will display as 

image

If the icon is used as it is, it leaves a nice padding between the text and icon.

Jokinen: Ah, correct. I forgot to check this change in the browser.

tuovinensanttu (approved):
Looking good!

PR #207, opened 2021-01-19
YM-396 | Fix cognitively challenging date input

Jokinen:
## Description

Using the number type would allow the arrow keys to be used in
selecting the content of a date part. This was however deemed too
confusing in the accessibility report.

## Context

[YM-396](https://helsinkisolutionoffice.atlassian.net/browse/YM-396)

## How Has This Been Tested?

I've tested this manually by using Safari and VoiceOver. There is no longer a mention of a spinner field, nor do the arrow keys work.

With a screen reader

1. Navigate to birthday input
2. Navigate through date input fields
3. Expect for fields to be described as text fields
4. Expect that up and down arrow don't change the value in the fields


tuovinensanttu (approved):
👍

PR #208, opened 2021-01-20
YM-397 | Use more descriptive error messages on forms

Jokinen:
## Description

Adds more descriptive error messages to forms which contain more details about what actions the user needs to take to remedy the situation.

Building on the previous assumptions, I attempted to create generic error messages that can be shared between fields. The upside is that there are fewer translation strings to maintain and fewer changes in the code, but the downside is that this approach entangles the translations with the code some more.

I would generally advise to use non-generic strings when the translation require more convoluted structures. I've applied field specific errors for instance to the email field.

## Context

[YM-397](https://helsinkisolutionoffice.atlassian.net/browse/YM-397)

## How Has This Been Tested?

I've tested this manually by going though the registration flow. I've added unit test to a new utility I added.

1. Go to edit information view
2. Test that three required fields get proper errors
3. Test that the first name field gives an error if you only input one character
4. Test that the school name gives an error if you input more than 10 characters
5. Test that the email field returns an error if you input an invalid email
6. Test that the postal code field returns an error when you input an invalid postal code

### DIFF
--- a/src/domain/youthProfile/youthProfileSchemas.ts
+++ b/src/domain/youthProfile/youthProfileSchemas.ts
@@ -9,9 +9,7 @@ import ageConstants from './constants/ageConstants';
 
 const isConsentRequired = (birthDate: string, schema: Yup.StringSchema) => {
   const userAge = differenceInYears(new Date(), new Date(birthDate));
-  return userAge < ageConstants.ADULT
-    ? schema.required('validation.required')
-    : schema;
+  return userAge < ageConstants.ADULT ? schema.required() : schema;
 };
 
 const requireIfNotAdult = (extendedSchema: Yup.StringSchema) => {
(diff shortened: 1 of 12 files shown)


src/domain/youthProfile/youthProfileSchemas.ts
```diff
@@ -89,19 +106,26 @@ const basicInformationSchema = Yup.object().shape({
...
 const additionalInformationSchema = Yup.object().shape({
-  schoolName: Yup.string().max(128, 'validation.tooLong'),
-  schoolClass: Yup.string().max(10, 'validation.tooLong'),
-  photoUsageApproved: Yup.string().required('validation.required'),
+  schoolName: Yup.string()
+    .max(128)
+    .label('fields.schoolName'),
+  schoolClass: Yup.string()
+    .max(10)
+    .label('fields.schoolClass'),
```
tuovinensanttu, line 125: There shouldn't be any scenario where these fields are required. But I guess it won't hurt to leave these here.

tuovinensanttu (commented):
For some reason `terms` field is not getting english translation 🤔 it works on finnish and swedish thought.

src/domain/youthProfile/youthProfileSchemas.ts
Jokinen: Did you comment the correct lines? AFAIK these two fields are not required.

Jokinen:
Seems that I've forgotten to add the translation into English or deleted it by accident which is why it's not visible.

src/domain/youthProfile/youthProfileSchemas.ts
tuovinensanttu: Exactly, maybe I miss understood something, but aren't these labels for displaying text "Required + field name"?

Jokinen: Ah, now I understand. The labels are also used for `min` and `max` errors. See here:

https://github.com/City-of-Helsinki/youth-membership-ui/pull/208/files#diff-85a7f9f12f8ccede1618ba1ef3c66c7b8cbb6da984cd2ef77af1400672330e31R215-R216

tuovinensanttu (approved):

Jokinen:
Rebased to resolve conflicts.

PR #209, opened 2021-01-20
YM-395 | Allow form fields to be autocompleted

Jokinen:
## Description

By adding autocomplete attributes, we allow users to autocomplete or -fill form fields. This makes the forms more accessible, because it may be swifter for users to input values with autofill.

https://www.w3.org/TR/WCAG21/#input-purposes

## Context

[YM-395](https://helsinkisolutionoffice.atlassian.net/browse/YM-395)

## How Has This Been Tested?

I've tested this manually by applying the attributes based on the resource linked above and checking that they can be found from elements.

I find this feature difficult to test, because it's unclear what logic browsers user to autofill fields. For instance Chrome would autofill fields correctly even without specifying these attributes.

In general terms, this procedure can be followed to check all the changes in case you are able to find a sandbox where you can test strictly autocomplete:

1. Navigate to login page
2. Expect birthday input to be autofillable with your birthday

.

1. Navigate to create/edit form
2. Expect your details (first name, last name, address fields, phone number, language) to be auto fillable
3. Expect approver fields not to be auto fillable

.

1. Navigate to approver field
2. Expect the approver fields to be auto fillable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 - [Accessibility] Links that did not warn when they opened in a new window
 - [Accessibility] Cognitively challenging date input
 - [Accessibility] Fixed the inconsistent use of HTML headings.
+- [Accessibility] Add support for autocomplete to form fields
 
 ## [1.2.1] - 2020-11-25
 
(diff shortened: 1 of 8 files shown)


tuovinensanttu (approved):
Couldn't test the registration form because all my accounts are already in use 😅  but looking good never the less!

PR #214, opened 2021-01-22
FIX | Change approver route to use AppPageTitleRoute

tuovinensanttu:
## Description

Changed approver route to use `AppPageTitleRoute` component. I forgot to change when reworking routing components.

## Context

-

## How Has This Been Tested?

Tested this manually, route works.

- Have a profile that has pending confirmation and navigate to approval route. It should work normally.


Jokinen (approved):
I think that this is more of a fix than a change. This change fixes the inaccessible approver view (with the current route the user's profile is checked, when they have none they are redirected into login).

tuovinensanttu:
> I think that this is more of a fix than a change. This change fixes the inaccessible approver view (with the current route the user's profile is checked, when they have none they are redirected into login).

I'll move it under fixed.

PR #215, opened 2021-01-22
Chore | Update Apollo from branch 2 to branch 3

Jokinen:
## Description

Updates Apollo from major version 2 into major version 3.

Apollo changed its package structure in conjunction with this change, which in turn caused some churn.

## How Has This Been Tested?

I've completed the registration flow for a new user and it worked. I've checked the test suite and linters.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,8 +4,7 @@
   "license": "MIT",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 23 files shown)


tuovinensanttu (approved):
Everything is working normally. LGTM.

PR #217, opened 2021-01-22
DEVOPS-228 - KuVa Tunnistamo to use

Hi-Fi:
## Description
Allows usage of Tunnistamo also at review environment(s)

## Context

[DEVOPS-228](https://helsinkisolutionoffice.atlassian.net/browse/DEVOPS-228)

## How Has This Been Tested?
Implemented to open-city-profile-ui, and tested also in created review env manually.

Access the review environment and sign in

### DIFF
--- a/.github/workflows/review.yml
+++ b/.github/workflows/review.yml
@@ -19,18 +19,19 @@ jobs:
     name: Build
     steps:
       - uses: actions/checkout@v2
+      - uses: andersinno/kolga-setup-action@v2
       - name: Build
         uses: andersinno/kolga-build-action@v2
         env:
           DOCKER_IMAGE_NAME: ${{ env.REPO_NAME }}-review
           DOCKER_BUILD_ARG_REACT_APP_ENVIRONMENT: 'review'
-          DOCKER_BUILD_ARG_REACT_APP_PROFILE_LINK: "https://profiili.test.kuva.hel.ninja/loginsso"
-          DOCKER_BUILD_ARG_REACT_APP_OIDC_AUTHORITY: "https://api.hel.fi/sso-test/"
-          DOCKER_BUILD_ARG_REACT_APP_OIDC_CLIENT_ID: "https://api.hel.fi/auth/jassari"
+          DOCKER_BUILD_ARG_REACT_APP_PROFILE_LINK: "https://profiili.${{ env.BASE_DOMAIN }}/loginsso"
+          DOCKER_BUILD_ARG_REACT_APP_OIDC_AUTHORITY: "https://tunnistamo.${{ env.BASE_DOMAIN }}/"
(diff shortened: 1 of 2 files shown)


.github/workflows/review.yml
```diff
@@ -24,13 +24,13 @@ jobs:
         env:
           DOCKER_IMAGE_NAME: ${{ env.REPO_NAME }}-review
           DOCKER_BUILD_ARG_REACT_APP_ENVIRONMENT: 'review'
-          DOCKER_BUILD_ARG_REACT_APP_PROFILE_LINK: "https://profiili.test.kuva.hel.ninja/loginsso"
-          DOCKER_BUILD_ARG_REACT_APP_OIDC_AUTHORITY: "https://api.hel.fi/sso-test/"
-          DOCKER_BUILD_ARG_REACT_APP_OIDC_CLIENT_ID: "https://api.hel.fi/auth/jassari"
+          DOCKER_BUILD_ARG_REACT_APP_PROFILE_LINK: "https://profiili.${{ env.BASE_DOMAIN }}/loginsso"
+          DOCKER_BUILD_ARG_REACT_APP_OIDC_AUTHORITY: "https://tunnistamo.${{ env.BASE_DOMAIN }}"
```
tuomas777, line 28: Don't know if it matters or not, but this seems to be missing a `/` at the end.

Hi-Fi: Probably matters, fixed.

.github/workflows/review.yml
```diff
@@ -45,3 +45,17 @@ jobs:
...
+          kubeconfig_raw: ${{ env.KUBECONFIG_RAW}}
+          review_namespace: ${{ env.K8S_NAMESPACE }}
+          api_scope: https://api.hel.fi/auth/jassariapi
+
+      - name: Setup Tunnistamo
+        uses: City-of-Helsinki/setup-tunnistamo-action@main
+        with:
+          kubeconfig_raw: ${{ env.KUBECONFIG_RAW}}
+          review_namespace: ${{ env.K8S_NAMESPACE }}
+          api_scope: https://api.hel.fi/auth/helsinkiprofile
```
charn, line 61: I guess it's not really possible in `setup-tunnistamo-action`, but I think the frontend would need two scopes `https://api.hel.fi/auth/jassariapi` and `https://api.hel.fi/auth/helsinkiprofile`? :thinking:

Hi-Fi: That's why the action is called twice, as one call is able to connect the client only to one scope. So the review client should be correctly having 2 scopes.

tuomas777 (approved):
LGTM 👍

.github/workflows/review.yml
charn: Ah, indeed, all good here then. :+1:

PR #218, opened 2021-01-22
Chore | Remove enzyme

Jokinen:
## Description

We upgraded to CRA 4 which adds more coupling with `testing-library`. We used both `testing-library` and `enzyme`. I refactored `enzyme` out of the application so that we aligned with CRA's direction.

## How Has This Been Tested?

I've ran the test suite and checked that it passes.


tuovinensanttu (approved):
All tests pass 👍

PR #220, opened 2021-01-27
Add react-toastify to change notification logic.

tuovinensanttu:
## Description

Installed `react-toastify` and wrapper the old `NotificationComponent` with it. 

## Context

This change allows us to get rid of the logic where `NotificationComponent` was imported to almost every component.

## How Has This Been Tested?
Tested this manually.

One way to test this is to use renew membership functionality.


src/domain/auth/components/login/Login.tsx
```diff
@@ -42,6 +42,10 @@ function Login(props: Props) {
 
   const isAuthenticated = useSelector(isAuthenticatedSelector);
 
+  if (Boolean(props.auth.error)) {
+    toastNotification({ onClose: () => props.resetError() });
+  }
```
tuovinensanttu, line 47: This solution seems to be working, only one notification is shown at the time. I'm still thinking if I should wrap this with `useEffect` to be sure.

CHANGELOG.md
```diff
@@ -36,6 +36,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 ### Changed
 
 - Use GitHub actions instead of Travis
+- centralize notification with react-toastify.
```
Jokinen, line 39: Minor, no capitalisation

src/common/components/notification/toastNotification.tsx
```diff
@@ -0,0 +1,23 @@
...
+};
+
+export type NotificationOptions = {
+  labelText?: string;
+  notificationMessage?: string;
+  type?: 'info' | 'success' | 'alert' | 'error';
+  onClose?: () => void;
+};
+
+const toastNotification = (notificationOptions: NotificationOptions) => {
```
Jokinen, line 20: This is stored under `components`. Is it really a component?

src/domain/app/App.tsx
```diff
@@ -19,7 +21,7 @@ import authConstants from '../../domain/auth/constants/authConstants';
 import authenticate from '../../domain/auth/authenticate';
 import AppMeta from './AppMeta';
 import AppRoutes from './AppRoutes';
-
+import styles from './app.module.css';
```
Jokinen, line 24: Missing whitespace.

src/domain/app/AppYouthProfileRoute.tsx
```diff
@@ -29,12 +29,9 @@ function AppYouthProfileRoute({
   ...rest
 }: Props) {
   const { t } = useTranslation();
-  const [showNotification, setShowNotification] = useState<boolean>(false);
 
   const { data, loading } = useQuery<HasYouthProfile>(HAS_YOUTH_PROFILE, {
-    onError: () => {
-      setShowNotification(true);
-    },
+    onError: () => toastNotification({}),
```
Jokinen, line 34: Using an empty object is discouraged.

src/domain/auth/components/login/Login.tsx
Jokinen: Well, you are correct. The render function should not have side effects unless wrapped in a `useEffect`. But I think that an even better route could be to set the toast in the authentication routine itself, in which case you would not need to use a `useEffect` either. Now the firing of the notification doesn't need to be tied to this component anymore.

Jokinen (commented):
Good changes! Added a few comments.

src/domain/app/AppYouthProfileRoute.tsx
tuovinensanttu: Fixed now. Silly mistake on my part to, forgot to make whole `notificationOptions` optional.

src/common/components/notification/toastNotification.tsx
tuovinensanttu: True! Moved it under helpers instead.

CHANGELOG.md
tuovinensanttu: Fixed!

tuovinensanttu:
> Well, you are correct. The render function should not have side effects unless wrapped in a useEffect. But I think that an even better route could be to set the toast in the authentication routine itself, in which case you would not need to use a useEffect either. Now the firing of the notification doesn't need to be tied to this component anymore.

Seems that I was so used to notification being a component that I didn't even think about this. Removed error handling from `Login` component and moved it to `authenticate` and `logout` methods.

Jokinen (approved):

PR #223, opened 2021-01-28
FIX | Fix browser tests

Jokinen:
## Description

One selector had fallen out of date since the HDS upgrade. It's now updated and browser tests should pass.

https://github.com/City-of-Helsinki/youth-membership-ui/actions/runs/518272513


tuovinensanttu (approved):
Tests pass.

PR #224, opened 2021-01-29
Fix apollo warning

tuovinensanttu:
Fixed warning `apollo` was throwing related to data not having identifier. Also made changes to `.env` and pointed pointed authentication to KuVa tunnistamo.

### DIFF
--- a/src/domain/membership/graphql/MembershipInformation.graphql
+++ b/src/domain/membership/graphql/MembershipInformation.graphql
@@ -2,6 +2,7 @@ query MembershipInformation {
     myYouthProfile {
         id
         profile {
+            id
             firstName
             lastName
         }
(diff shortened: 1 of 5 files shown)


src/domain/membership/graphql/MembershipInformation.graphql
```diff
@@ -1,7 +1,7 @@
 query MembershipInformation {
     myYouthProfile {
+        id
         profile {
-            id
```
Jokinen, line 4: Seems like this `id` is used somewhere in the code.

Jokinen (changes requested):
Tests or the build do not pass.

src/domain/membership/graphql/MembershipInformation.graphql
tuovinensanttu: It is indeed. I thought I double checked everything, after taking another look it was used in the QRcode that points to admin. Added it back.

Jokinen (approved):

PR #227, opened 2021-02-01
YM-450 | Fix production build

Jokinen:
## Description

Downrgades the production build configuration so that we can rebuild
the current production version which assumes that Jässäri has not yet
been separated from Helsinki profile.

After merging these changes, we can re-release production and it'll be built with config that are compatible with its vurrent version.

## Context

[YM-450](https://helsinkisolutionoffice.atlassian.net/browse/YM-450)


tuovinensanttu (approved):
As far as I remember, these are correct.

PR #231, opened 2021-02-02
HEL-403 | Fix terms heading

tuovinensanttu:
A quick fix that changes `terms of service` title from `h3` to `h2` to make it inline with rest of the form.

### DIFF
--- a/src/domain/youthProfile/form/YouthProfileForm.tsx
+++ b/src/domain/youthProfile/form/YouthProfileForm.tsx
@@ -145,7 +145,9 @@ function YouthProfileForm(componentProps: Props) {
             <PageSection>
               {!componentProps.isEditing && (
                 <React.Fragment>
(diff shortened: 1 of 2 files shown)


Jokinen:
I'm thinking whether we should extend the API of the `Text` component or not? We could add a `component` prop which developers could then use to force the element that's used to render--while using the styles of some variant. What do you think?

tuovinensanttu:
> I'm thinking whether we should extend the API of the Text component or not? We could add a component prop which developers could then use to force the element that's used to render the component--while using the styles of some variant which by default assumes a certain element. For instance <Text component="h2" variant="h3" /> would result in <h2 class="h3">.... Would it be confusing? What do you think?

While checking the elements it might seem little confusing, but there is quite a few cases where additional classname is passed to reduce the size of the text. I think this would be a good idea.

Jokinen (approved):

PR #242, opened 2021-02-18
Release 1.3.0

charn:
This PR releases the front-end which will use the new separated youth-membership backend.


Jokinen (approved):
Added documentation.

I checked the prod documentation and as far as I can see they look correct.

PR #246, opened 2021-02-18
Release 1.3.2

charn:
`https://api.hel.fi/auth/jassariapiprod` should be used instead of `https://api.hel.fi/auth/jassariapi`


Jokinen (approved):
Looks good 👍

PR #247, opened 2021-02-24
YM-466 | Fix photo usage always being set as true

Jokinen:
## Description

Fixes the photo usage approval boolean check.

It's a bit weird that a string value is used for the approval value, but I made the most minor change possible and just fixed the check.

## Context

[YM-466](https://helsinkisolutionoffice.atlassian.net/browse/YM-466)

Ensure that >=15 && <18 year old can deny photo usage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 
 ## Unreleased
 
+### Fixed
+
+- Approval process always allowing photo usage regardless of choice
+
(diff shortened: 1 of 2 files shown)


tuovinensanttu (approved):
👍 . If I'm not mistaken, photoUsage has string value because we are using radio buttons, which work really poorly / not at all with boolean values.

Jokinen:
Ok, good to know. The difficulty may be in part due to using a radio group for a yes/no choice that by default would be better handled with a checkbox. In any case, in this scenario, it could still be for the best to normalize the value into a boolean value before saving it to a store, or at least before sending it out of the form.

PR #248, opened 2021-02-26
YM-398 | Announces title changes with an aria-live region

Jokinen:
## Description

Announces title changes with an aria-live region.

I mostly copy pasted the implementation from kukkuu. Unfortunately I could not come up with a clear destination for this code.

## Context

[YM-398](https://helsinkisolutionoffice.atlassian.net/browse/YM-398)


tuovinensanttu (approved):
Tested this again with voiceOver + NVDA and the titles are read aloud.

PR #249, opened 2021-03-02
YM-469 | Reflect membership status changes in membership info

Jokinen:
## Description

Previously the membership status of RENEWING was not reflected in the UI. It was also unclear which of the other states (ACTIVE, PENDING, EXPIRED) were handled to what degree.

As added complexity, we also want to show a slightly different UI for user's who are "adults" as opposed to when they are minors and need to have their membership approved.

As context, here are the different possible states:

**Membership status**  
PENDING -- Membership is not active and is awaiting approval from an approver  
ACTIVE -- Membership is active  
RENEWING -- Membership is active and is awaiting approval from an approver  
EXPIRED -- Membership has expired  

**On top of the status there other fields are used**  
`renewable` -- True when the membership can be renewed. True after `1.5` and until the membership is renewed.  
`birthDate` -- User's birthday, used to determine whether the user is a minor.  
`approvedTime` -- The time the user's membership has been approved. Missing for users who have just registered.

The relevant design files can be found from here: https://app.abstract.com/projects/b80d4d50-787f-11e9-ae1d-ef140a99b44b/branches/master/collections/e26374e2-c8b6-43a0-b017-3b6bf8d9e7b9

## Context

[YM-469](https://helsinkisolutionoffice.atlassian.net/browse/YM-469)

## How Has This Been Tested?

I've added unit tests to the membership view that validate against the design's expectations.

This is quite difficult to test as all of the states can't be reproduced by editing data from the admin. I tested the states locally by hardcoding values. The behaviour should be in line with the design. Moreover, the current view the user is redirected to after registration should be unchanged.

### DIFF
--- a/src/domain/membership/information/MembershipInformation.tsx
+++ b/src/domain/membership/information/MembershipInformation.tsx
@@ -6,8 +6,8 @@ import { Button } from 'hds-react';
 import { MembershipInformation as MembershipInformationTypes } from '../../../graphql/generatedTypes';
 import LinkButton from '../../../common/components/linkButton/LinkButton';
 import convertDateToLocale from '../../../common/helpers/convertDateToLocale';
-import getFullName from '../helpers/getFullName';
 import MembershipPageLayout from '../MembershipPageLayout';
+import Membership from '../MembershipUtility';
 import styles from './membershipInformation.module.css';
 
 interface Props {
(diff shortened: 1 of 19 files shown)


src/domain/membership/information/MembershipInformation.tsx
```diff
@@ -3,75 +3,316 @@ import { useTranslation } from 'react-i18next';
...
+  memberShipStatus,
+  membershipRenewable,
+}: MembershipExpiryTitleProps) {
+  const { t } = useTranslation();
+
+  const isRenewable = membershipRenewable;
+  const isExpired = Membership.getIsExpired(memberShipStatus);
+  const isActive = Membership.getIsActive(memberShipStatus);
+  const isRenewing = Membership.getIsRenewing(memberShipStatus);
+  const isPending = Membership.getIsPending(memberShipStatus);
```
tuovinensanttu, line 33: These variables are declared four times in this file. I'd consider combining them under one getter function e.g `Membership.getStatus` that would return an object containing all the values. This could then be destructured to get the variables `isExpired, isActive` etc.

tuovinensanttu (approved):
Tested the UI and it seems to be working nicely. There were quite a lot of changes and they indeed clear up the application flow. Nice work!

src/domain/membership/information/MembershipInformation.tsx
Jokinen: I agree that the API would be nicer for this usecase in that there wouldn't be as many lines. However, I'm hesitant to create the extra level of abstraction, because I am not sure about how well it would fare with other usecases. At this point I feel like the change would be mostly syntax sugar as there would still be as many declarations, but with the chance that it'd be inconvenient in some other context later on.

However, I'm willing to make this change if there's some perspective I'm not noticing.

tuovinensanttu: True, amount of declarations would be the same, but imo would clean up syntax a little bit and declaring them one by one feels kind of "heavy". We can leave it as it is for now, but it might be worth it the future if we implement other features that rely on multiple statuses.

PR #250, opened 2021-03-04
YM-257, YM-261 | Add language to url

Jokinen:
## Context

[YM-257](https://helsinkisolutionoffice.atlassian.net/browse/YM-257)
[YM-261](https://helsinkisolutionoffice.atlassian.net/browse/YM-261)


tuovinensanttu (approved):
Sorry, took me a while to go through this one. Nothing to comment on, I think this solution makes sense and is implemented really well!

PR #251, opened 2021-03-05
YM-469 | Add custom success message for minors

Jokinen:
## Description

Add a custom success message for minors which tells them the renewal request has been sent for approval.

## Context

[YM-469](https://helsinkisolutionoffice.atlassian.net/browse/YM-469)

### DIFF
--- a/src/domain/membership/MembershipUtility.ts
+++ b/src/domain/membership/MembershipUtility.ts
@@ -4,50 +4,48 @@ import { MembershipStatus } from '../../graphql/generatedTypes';
 import ageConstants from '../youthProfile/constants/ageConstants';
 import Config from '../../config';
 
-class Membership {
-  static getAdminUrl(profileId: null | string) {
-    const url = Config.adminUrl;
+export function getAdminUrl(profileId: null | string) {
+  const url = Config.adminUrl;
 
-    if (!profileId) {
-      return null;
-    }
-
-    return `${url}youthProfiles/${profileId}/show`;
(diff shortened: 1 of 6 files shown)


tuovinensanttu (approved):
Tested the toast with minor profile and correct text was shown. I'm not sure if I remember this correctly, but isn't guardian supposed to approve the renewal before the QR-code is shown again? Anyway, seems to be a backend problem.

Jokinen:
Hmn, I can try to find out if something not working as it should. Could you write down in more detail how you tested this?

Here's my assumption

With an existing active account with a birthday that makes the account a minor

1. Go to Django admin and set expiration date into 2019
2. Access UI  and renew membership
3. See Qr code unexpectedly

tuovinensanttu:
Thats correct, sorry, I could have written more specific steps in the first place. 

I have a minor account with birthdate set to 1.1.2006.

So after pressing the renew, and seeing the QR-code, I checked the django admin and expiration date was updated to 2021-08-31. Because the profile status isn't shown on django -admin, it's hard to say that if that's set to active as well.

Jokinen:
I'll check this by checking what is returned in the API call and go from there. I'll be in contact with BE if I need to.

Jokinen:
I think that this behaviour is just due to the BE inferring that the current approveTime applies to the ongoing membership term. I'll verify with the BE.

Jokinen:
Confirmed that this is known behaviour 👍

PR #252, opened 2021-03-05
YM-478 | Remove user survey feedback link

Jokinen:
## Description

User survey is over so we can remove the link to it.

## Context

[YM-478](https://helsinkisolutionoffice.atlassian.net/browse/YM-478)

## How Has This Been Tested?

I've manually verified that the link is missing.


tuovinensanttu (approved):
👍

PR #259, opened 2021-03-11
YM-487 | Option to define additional locales with environment variables

Jokinen:
## Description

Allows additional locales to be defined with `REACT_APP_ADDITIONAL_LOCALES` environment variable.

The format for the config is a string of comma separated ISO 639-1 values.

Locales found from the config are used to search for translation files, which are then configured for i18next.

Additional languages have been configured for use in review and staging environments. Users should expect to see error for missing locales.

## Context

[YM-487](https://helsinkisolutionoffice.atlassian.net/browse/YM-487)

## How Has This Been Tested?

I've added or modified unit level tests.

### DIFF
--- a/src/i18n/I18nService.ts
+++ b/src/i18n/I18nService.ts
@@ -4,35 +4,70 @@ import LanguageDetector from 'i18next-browser-languagedetector';
 import * as H from 'history';
 
 import * as PathUtils from '../common/reactRouterWithLanguageSupport/pathUtils';
+import Config from '../config';
+import Logger from '../logger';
 import en from './en.json';
 import fi from './fi.json';
 import sv from './sv.json';
 import setYupLocale from './setYupLocale';
 
-const supportedLanguages = ['fi', 'sv', 'en'];
-export type Language = typeof supportedLanguages[number];
+const defaultLanguages = ['fi', 'sv', 'en'];
+export type Language = string;
+
+function getLanguage(languageCode: string) {
+  let resource;
+
+  try {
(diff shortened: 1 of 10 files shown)


src/i18n/I18nService.ts
```diff
@@ -4,35 +4,70 @@ import LanguageDetector from 'i18next-browser-languagedetector';
...
+import Logger from '../logger';
 import en from './en.json';
 import fi from './fi.json';
 import sv from './sv.json';
 import setYupLocale from './setYupLocale';
 
-const supportedLanguages = ['fi', 'sv', 'en'];
-export type Language = typeof supportedLanguages[number];
+const defaultLanguages = ['fi', 'sv', 'en'];
+export type Language = string;
```
Jokinen, line 15: I had to use a more general type because we don't have knowledge about which languages may be configured.

tuovinensanttu (approved):
Looking good, nice way to add additional languages to application. Changing the language from url to something that doesn't have translations defaults to finnish nicely, this breaks the Accessibility statement view though.

EDIT: Not like break break, it just says invalid language which is understandable :)

Jokinen:
In general, I don't mind if the application breaks when it has been configured to work with a locale, but is missing required definitions for that locale. In that case the application is misconfigured as far as I am concerned. I thought about ignoring locales when there are missing translations, but decided that it's more transparent to just fail and display errors.

But your point about the accessibility document is very good. I had forgotten that it uses a very custom system to provide translation support. The current approach will be difficult to maintain with eight languages. I'll try to account for the accessibility documents while I'm developing the rest of the extended language support related features.

PR #261, opened 2021-03-12
YM-489 | Use i18n service to find options for language menu

Jokinen:
## Description

Instead of using hardcoded values, the navigation uses the I18nService to find out which languages are supported.

Also adds language direction support into the language navigation options.

## Context

[YM-489](https://helsinkisolutionoffice.atlassian.net/browse/YM-489)


tuovinensanttu (approved):
👍

PR #270, opened 2021-03-17
YM-498 | Allow approver to change language at home value

Jokinen:
## Description

Due to the whitelisting logic, changes to the language at home field int he approver field did not take effect.

I also adjusted the additional contact languages in this PR. They need to now be toggled on with a feature flag. This is to give better control over when we want to release additional langauges.

## Context

[YM-498](https://helsinkisolutionoffice.atlassian.net/browse/YM-498)


tuovinensanttu (approved):
LGTM 👍

PR #272, opened 2021-03-24
YM-479 | Add email template resource files

Jokinen:
By serving these files through the UI app, they get served with nginx
without much extra configuration.

## Description

Adds image resources that are used by the email templates into the application.

We decided to serve the files through this application because it's served with nginx. If we were to use the default approach provided by Django, we would need additional configuration in order to enable nginx.

## Context

[YM-479](https://helsinkisolutionoffice.atlassian.net/browse/YM-479)


charn (approved):
LGTM! :+1: :shipit:

PR #279, opened 2021-05-24
YM-494 | Add and update translations

charn:
This PR adds/updates the translation files for the new additional languges: ar, et, fr, ru, so.

Refs YM-494


Jokinen (approved):
I gave this a quick test locally and it looked good.

Browser tests are not passing for some reason. Based on the screeenshot some API call fails. It may very well be unrelated to this change 🤔

charn:
To get the tests passing I needed to add the client id `youth-membership-ui-feature-ym-494-279` under the youth membership service in the open-city-profile staging instance. So, it looks like this change doesn't break the test, but basically all new PR will fail the browser test unless the client id is added in a similar way.

PR #288, opened 2021-11-29
YM-512 | Reorder GraphQL calls for creating youth membership

charn:
## Description
Due to open-city-profile deprecating some old behavior, the order of the calls is changed when registering a new member to:

1. Create Helsinki profile
2. Add connection to Youth membership service
3. Create Youth profile

The last call eventually also calls the Helsinki profile in the youth membership backend. The service connection needs to be created into Helsinki profile before the youth membership profile creation call is made.

## Context
[YM-512](https://helsinkisolutionoffice.atlassian.net/browse/YM-512) Change the order in which the service connection to Youth membership is made

## How Has This Been Tested?
I used the latest development branch of open-city-profile, which doesn't have the implicit connection feature/logic anymore, which allowed the old way of creating the youth membership profile before the service connection is created into Helsinki profile. 

I checked that without the changes the youth-membership backend was unable to successfully create the youth profile (which requires querying the Helsinki profile backend). With the changes there's no errors.

### DIFF
--- a/src/domain/youthProfile/hooks/useCreateProfiles.ts
+++ b/src/domain/youthProfile/hooks/useCreateProfiles.ts
@@ -48,11 +48,7 @@ const useCreateProfiles = ({ onError }: Options) => {
     addServiceConnection,
     { loading: addingServiceConnection },
   ] = useMutation<AddServiceConnectionData, AddServiceConnectionVariables>(
-    ADD_SERVICE_CONNECTION,
-    {
-      refetchQueries: ['HasYouthProfile', 'NameQuery'],
-      awaitRefetchQueries: true,
-    }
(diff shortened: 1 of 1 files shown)


Jokinen (commented):
Codewise this looks good to me. Federation was down when I was testing so could not yet try this out in the UI, or verify if the browser tests are really broken.

Jokinen (approved):
I restarted the federation and renewed the browser test user's profile.

After restarting federation, I was able to test this user path.

I deleted my existing accounts through the admins as kuva's profiili UI no longer works.

I then created a new account for myself. The account was created successfully.

I was also able to successfully run these tests locally. Wasn't able to determine why the browser tests are not passing in this PR.

PR #305, opened 2022-10-12
Removed the update and remove addresses fields from the create profile mutation

nikomakela:
YM-515. Production hotfix. Updated the schema by generating the new types with codegen.

Testing instruction:
1. Go to https://jassari.test.kuva.hel.ninja
2. Create a profile (this failed before the fix and it was shown in the network tab and with a visible notification)
3. Update a profile (this uses the same getAddress -functions as the creation, so needs a test also)

### DIFF
--- a/src/domain/youthProfile/approve/__test__/__snapshots__/ApproveYouthProfileForm.test.tsx.snap
+++ b/src/domain/youthProfile/approve/__test__/__snapshots__/ApproveYouthProfileForm.test.tsx.snap
@@ -57,7 +57,7 @@ exports[`matches snapshot 1`] = `
           <span
             class="value"
           >
-            01.01.2007
+            01.01.2008
           </span>
         </div>
         <div
--- a/src/domain/youthProfile/helpers/contactInformationUtils.ts
+++ b/src/domain/youthProfile/helpers/contactInformationUtils.ts
@@ -1,0 +1,123 @@
+import { isEqual } from 'lodash';
+
+import {
+  PrefillRegistartion,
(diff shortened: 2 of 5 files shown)


karisal-anders (commented):
It would be easier to code review this if the moving around of code from createProfileMutationVariables.ts to contactInformationUtils.ts was not combined with functional changes in a single commit. Now one has to go through the code and see if something has functionally changed or was it just moved from a file to another. And yes, I see this is a hotfix so you might be in a hurry but something hopefully to keep in mind when doing further code changes. And also yes, I haven't always abided by this myself. But, definitely good for code review IMHO!

jakezu:
-

jakezu (commented):
Test

jakezu (approved):

nikomakela:
> It would be easier to code review this if the moving around of code from createProfileMutationVariables.ts to contactInformationUtils.ts was not combined with functional changes in a single commit. Now one has to go through the code and see if something has functionally changed or was it just moved from a file to another. And yes, I see this is a hotfix so you might be in a hurry but something hopefully to keep in mind when doing further code changes. And also yes, I haven't always abided by this myself. But, definitely good for code review IMHO!

But I kind of needed to do it, because I only wanted to remove those fields from CreateProfileMutation and not from the UpdateProfileMutation. It's in its own commit anyway. It was a bad thing that the update used the util-function from createMutation - it should always have been in it's own utility file. 

There are not so many lines of code anyway 😛 

@karisal-anders Is the change approved or not?

